PR #181, opened 2013-08-03
Remove `merge.ff=only` setting

gylaz:
- add `git ff` alias that does the same thing

It's not possible to do a non-ff merge when `merge.ff=only` is set.
There are certain situations (like merging from upsteam repo) when
merging without ff is desirable.

### DIFF
--- a/gitconfig
+++ b/gitconfig
@@ -10,6 +10,7 @@
   co = checkout
   create-branch = !sh -c 'git push origin HEAD:refs/heads/$1 && git fetch origin && git branch --track $1 origin/$1 && cd . && git checkout $1' -
   delete-branch = !sh -c 'git push origin :refs/heads/$1 && git remote prune origin && git branch -D $1' -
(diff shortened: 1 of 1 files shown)


dcalhoun:
@jferris does the `--no-ff-only` flag still exist? When I attempt to use that flag it get: `error: unknown option 'no-ff-only'`. I'm using git v2.7.1.

Granted, I know that this repo primarily exists to help keep the thoughtbot team aligned on practices, but the `ff-only` config is incredibly limiting for outside users like myself. The inability to circumvent it (without deleting the lines from gitconfig) is a bit painful.

bernerdschaefer:
@dcalhoun If your fast-forward policy differs from that set in the dotfiles, you should be able to set `merge.ff=true` or `merge.ff=false` in `~/.gitconfig.local` to override it.

dcalhoun:
Thanks, @bernerdschaefer. I'll that a try.

PR #333, opened 2015-01-13
Remove outdated references to `Rnavcommand`

geoffharcourt:
`Rnavcommand` has been removed from rails.vim. These commands don't work if
you are using a recent version of the plugin. Users who want to regain this
functionality can do it through projections.


JoelQ:
:+1:

gylaz:
Thank you @geoffharcourt.

PR #335, opened 2015-01-26
Add `git rename-branch` alias

croaky:
* Extract `git current-branch` alias.
* Re-use `git current-branch` in two places.


gylaz:
LGTM

PR #337, opened 2015-02-03
Don't interpolate environment variables in aliases

reshleman:
A few aliases contain references to environment variables, but were
defined using double quotes. This caused zsh to interpolate the value of
those variables when the alias was defined instead of when it was
executed. In particular, any change to `PATH` (or `EDITOR` or `VISUAL`)
in `.zshrc.local`, which is sourced after `.aliases`, would not be
reflected in these aliases.

This commit defines these aliases using single quotes so that the
environment variables are evaluated when the alias is executed.


gylaz:
Thanks @reshleman!

PR #338, opened 2015-02-04
Stop ag from excluding directories with ds_store

danielnolan:
* Silver searcher or ag was filtering out any directory that had a
  .DS_STORE file. Any directory that has a .DS_STORE file in it is being
  ignored by ag and therefor does not show up in the ctrlp fuzzy finder
  either.
* Removing the * from before .DS_STORE in the gitignore
  file still prevents .DS_STORE files from being checked into source
  control and stops ag from filtering out the directory. I am not sure
  if this is a bug with silver_searcher or what

### DIFF
--- a/gitignore
+++ b/gitignore
@@ -1,4 +1,4 @@
-*.DS_Store
+.DS_Store
 *.sw[nop]
 .bundle
 .env


danielnolan:
this is happening with the most current version of silver_searcher from brew. ag version 0.29.0. I had @brandoncordell verify that he is experiencing the same issue.

brandoncordell:
I did experience the same issue and can confirm @danielnolan's commit fixes the issue for me.

jferris:
@gylaz any objection to merging this in? Seems sane to me.

geoffharcourt:
After experiencing this problem this morning, I dropped back to v 0.28.0 of Ag and the problem went away. I've filed an Ag issue [here](https://github.com/ggreer/the_silver_searcher/issues/598).

gylaz:
I think this makes sense to merge since specifying just `.DS_Store` makes more sense. I'm curious why it was `*.DS_Store` to begin with. Does OSX ever name these files with something before the `.`?

geoffharcourt:
The [Ag issue](https://github.com/ggreer/the_silver_searcher/issues/598) that caused the files to be incorrectly filtered out has been resolved in Ag v0.29.1.

danielnolan:
yes this was caused by a regression in Ag v0.29.0, but I agree with @gylaz why was it ever *.DS_STORE to begin with?

geoffharcourt:
@danielnolan, I agree as well that the asterisk should be removed.

gylaz:
Merged as 18a4c817bb237ce875ab92d7c9535b331e42e562. Thanks ya'll!

Issue #339, opened 2015-02-10
Reverse history search breaks cursor position


nickpad:
I've noticed that #335 looks like it's introduced a problem when using `Ctrl-r` to do a reverse history search in zsh. The cursor moves back into the prompt when in a git repository:

![reverse-search](https://cloud.githubusercontent.com/assets/588/6125702/1e56cebc-b16d-11e4-9723-cf6f65da89df.png)

The problem looks to have been introduced in the change to `zshrc` in [this commit](https://github.com/thoughtbot/dotfiles/commit/b0fe7c01ff24b40fe64736c1d5b2ac9d0f33477c).

I've spent a bit of time playing around with the prompt string but unfortunately haven't been able to work out where the problem lies yet.

geoffharcourt:
I have the same issue. I had thought it was something wrong with my terminal setup.

gylaz:
#340 has been merged. Can you try and see if latest on master fixes this?

nickpad:
Can confirm that it's now fixed for me. Thanks!

PR #340, opened 2015-02-12
Prevent unreadable prompt within Git repository

lunohodov:
When within a git repository, doing a tab-autocomplete on a command results with the command becoming unreadable. The command still works but this is pretty annoying visually.

Basically I

1. Navigate to a directory with a git repository in it. The prompt indicates the current branch properly
2. Next, type `git -` and hit tab

The prompt now shows only part of branch's name with the first suggestion appended.

zsh-prompt-unreadable

After googling a bit I stumbled upon several pages describing a similar problem (most useful to me was http://stackoverflow.com/questions/23740862/issues-with-zsh-prompt).

The culprit seems to be the git_prompt_info function escaping the `$current_branch` variable as if it is a color. As a result zsh is confused where the cursor is. After "unescaping" the variable everything seems to work fine.

Using zsh 5.0.7 (x86_64-apple-darwin13.4.0).

### DIFF
--- a/zshrc
+++ b/zshrc
@@ -2,7 +2,7 @@
 git_prompt_info() {
   current_branch=$(git current-branch 2> /dev/null)
   if [[ -n $current_branch ]]; then
-    echo " %{$fg_bold[green]%}%{$current_branch%}%{$reset_color%}"
+    echo " %{$fg_bold[green]%}$current_branch%{$reset_color%}"
   fi
 }
 setopt promptsubst


derekprior:
I have this problem as well. This change does fix it.

derekprior:
See also #339

croaky:
Oh, nice. I've seen this is as well but didn't know the cause. Thanks, @lunohodov, for fixing it! @gylaz, good to merge?

gylaz:
Merged. Thank you, @lunohodov!

PR #344, opened 2015-02-18
bin/replace improvement

NewAlexandria:
for those with an `.ackrc` that sets `--color`, this script will fail.

Set the `--nocolor` flag to make piping `xargs` safe

### DIFF
--- a/bin/replace
+++ b/bin/replace
@@ -9,4 +9,4 @@ shift
 replace_with=$1
 shift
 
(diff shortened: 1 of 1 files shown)


Keithbsmiley:
This sounds good  to me!

gylaz:
Thanks @NewAlexandria! Merged as 44f678312da32795e5443c973a10de77acf57948.

PR #346, opened 2015-02-20
Add alias for pull request with last commit msg

danielnolan:
* When opening a pull request on Github using the web interface
  the last commit subject and description are used for the pull request
  title and description. When using hub from the command line to open a
  pull request you have to specify the message using the -m flag,
  or you can have it read from a file or stdin using -F flag.
* Add alias pr-lmsg to pipe the last commit message from the current
  branch to hub for creating pull requests using the last git commit
  message.

### DIFF
--- a/gitconfig
+++ b/gitconfig
@@ -17,6 +17,7 @@
   delete-branch = !sh -c 'git push origin :refs/heads/$1 && git branch -D $1' -
   merge-branch = !git checkout master && git merge @{-1}
   pr = !hub pull-request
+  pr-lmsg = !git log -n 1 --pretty=format:'%s%n%n%b' | hub pull-request -F -
   rename-branch = !sh -c 'old=$(git current-branch) && git branch -m $old $1 && git push origin --set-upstream $1 && git push origin --delete $old' -
   st = status
   up = !git fetch origin && git rebase origin/master


adarsh:
I like this.

danielnolan:
@adarsh thanks. can't beat `g push && g pr-lmsg` to open a pull request with title and body from the last commit.

jessieay:
This is cool. 

One thought: in what case would you be opening a pull request with multiple commits? Ideally a pull request is small and only requires a single commit. In that case, that `hub` would use the commit message from the single commit for the PR description. 

If it is true that an ideal PR is small and only has one commit, does this change make it easier for us to make large / messy PRs in the future?

danielnolan:
@jessieay This isn't to open a pull request with multiple commits. Running `hub pull-requst` does not use the last commit message, it prompts you to enter a title and description for the pull-request.

jessieay:
@danielnolan are you sure? When I just tried it on a branch with one commit, this is what it pulled up:

screen shot 2015-03-18 at 12 15 12 pm

So I have the option of changing it, but the default behavior is to use last commit's message / description

danielnolan:
@jessieay interesting. This is what I get, and there is only one commit on that branch. I even rebased it to be sure. 
screen shot 2015-03-18 at 3 23 02 pm

What version of hub are you running? I just tried this on two machines both running hub 2.2.0 and I get the same result. 

Regardless this is still a shortcut since you can't open a pull request with hub without it opening the editor. If it's using the last message, there is no reason for the editor to open just so I can close it.

brandoncordell:
I'm running `hub 2.2.0` as well. This is the result I get.

```
# Requesting a pull to brandoncordell:master from brandoncordell:test-pr
#
# Write a message for this pull request. The first block
# of text is the title and the rest is description.
```

jessieay:
Was running 1.10 but upgraded to 2.2 and am getting the same results as before. Weird! If that is not default, I :+1: this PR - now to figure out why I am seeing this behavior...

danielnolan:
@jessieay I verified with the developer of hub that you are correct and if only one commit then hub should populate the editor with the last commit title and body to be used for the pull-request title and body. I am still having issues with it on my end however. https://github.com/github/hub/issues/847. I also found another issue that was opened on hub requesting a way to open a pull-request with the last-commit message without hub opening the editor https://github.com/github/hub/issues/722, but the issue was closed as will not implement. I am going to close this pr in favor of this one https://github.com/thoughtbot/dotfiles/pull/362

jessieay:
@danielnolan nice!

PR #348, opened 2015-02-20
Add `regreplop.vim`

iwz:
See: https://github.com/vim-scripts/regreplop.vim

* Type `<C-[>` followed by a motion to replace (paste over).
* Type `<C-[><C-[>` to paste over the entire line
* Visually select some text, then type `<C-[>` to paste over that text

For example:

Given you have "foobar" copied into your paste register
And your caret is on the `u` in this string: "fizzbuzz"
When you type `<C-[>iw`
Then the string will change from "fizzbuzz" to "foobar"

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -142,6 +142,12 @@ nnoremap <C-k> <C-w>k
 nnoremap <C-h> <C-w>h
 nnoremap <C-l> <C-w>l
 
+" Remap regreplop's replace operator
+" to not conflict with quick window movement's <C-k>
+nmap <C-[> <Plug>ReplaceMotion
+nmap <C-[><C-[> <Plug>ReplaceLine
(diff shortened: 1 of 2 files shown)


gylaz:
Is this the same as visual-mode highlighting then pressing `p`?

I can do this without plugin:
```
Given you have "foobar" copied into your paste register
And your caret is on the u in this string: "fizzbuzz"
When you type viwp
Then the string will change from "fizzbuzz" to "foobar"
```

iwz:
@gylaz the difference is that `viwp` yanks into your register, while `regreoplop` does not. further, `viwp` would not be repeatable.

iwz:
Closing in favor of https://github.com/thoughtbot/dotfiles/pull/351

PR #350, opened 2015-02-23
Add the `-v` flag to `commit --amend`.

blainesch:
This allows us to always see things we are going to commit, even if we
are amending.


jferris:
Seems good to me.

blainesch:
Re-pushed. I didn't change code, just who I committed as, it didn't have my correct email.

What's the correct way to version the .local files?

blainesch:
Rebased.

iwz:
:+1:

gylaz:
Thanks @blainesch!

croaky:
Oh good call.

PR #351, opened 2015-02-23
Add ReplaceWithRegister, ReplaceWithSameIndentRegister

iwz:
See: https://github.com/vim-scripts/ReplaceWithRegister
See: https://github.com/vim-scripts/ReplaceWithSameIndentRegister

```
USAGE
[count]["x]gr{motion}   Replace {motion} text with the contents of register x.
                        Especially when using the unnamed register, this is
                        quicker than "_d{motion}P or "_c{motion}<C-R>"
[count]["x]grr          Replace [count] lines with the contents of register x.
                        To replace from the cursor position to the end of the
                        line use ["x]gr$
{Visual}["x]gr          Replace the selection with the contents of register x.

[count]["x]grR          Replace [count] lines with the contents of register x
                        and adjust the indent to the first replaced line (like
                        pasting with ]p).
{Visual}["x]gR          Replace the selection with the contents of register x
                        and adjust the indent to the first selected line.
                        As the replacement is always entire lines, this only
                        works with linewise selections (V).
```

```cucumber
Given you have "foobar" copied into your paste register
And your caret is on the u in this string: "fizzbuzz"
When you type griw
Then the string will change from "fizzbuzz" to "foobar"
And the register will not be changed to "fizzbuzz", remaining "foobar"
And pressing . to repeat the command will work on another word
```


iwz:
A nice thing about this is `gR` which will paste and correct indentation

iwz:
Any thoughts on this one @gylaz?

gylaz:
I don't have any objections.

gylaz:
@iwz Has this still been useful to you? Do you think it's worth adding to dotfiles for everyone? Or is this more of a personal preference kind of thing? I want to avoid adding things to dotfiles that are be useful to only a handful of us, but not for everyone. Thoughts?

iwz:
@gylaz Yes, I use it every day. I do think it's useful for everyone, as equally useful as `vim-surround`. It's one of those things that once people understand they really like and always want to remember (at least with those I've showed this to).

Totally fine with leaving it out and adding it to my own `.local` though.

gylaz:
@iwz I'm hesitant to merge this, as it seems like not a lot of our dotfiles users will use this (hard to say). I really don't want to add everything that we find useful to our dotfiles, to prevent it from becoming a dropbag. The concern here is that this is a plugin, which will affect the boot of vim, and it's value may not be warranted (if not heavily used).

gylaz:
@iwz I'm going to close this. I feel like this lands in the `.local` land. Feel free to re-open if you disagree.

PR #352, opened 2015-02-24
Add vim-polyglot to plugins

tsigo:
Found this neat plugin the other day that includes many language-specific Vim
plugins so you don't have to add them individually as you start playing with a
new language.

[Check out the full list in the README](https://github.com/sheerun/vim-polyglot#language-packs).


vimrc
```diff
@@ -39,10 +39,6 @@ augroup vimrcEx
     \   exe "normal g`\"" |
     \ endif
 
-  " Set syntax highlighting for specific file types
-  autocmd BufRead,BufNewFile Appraisals set filetype=ruby
```
tsigo: https://github.com/sheerun/vim-polyglot/blob/master/ftdetect/polyglot.vim#L178

vimrc
```diff
@@ -39,10 +39,6 @@ augroup vimrcEx
...
-  autocmd BufRead,BufNewFile *.md set filetype=markdown
```
tsigo: https://github.com/sheerun/vim-polyglot/blob/master/ftdetect/polyglot.vim#L102-L107

Keithbsmiley:
This plugin is super heavy weight. Personally I'd really rather not pull it in.

Keithbsmiley:
Also many syntax files aren’t kept up to date. See https://github.com/sheerun/rspec.vim Which is a fork of my repo which hasn't pulled in changes for over a year.

gylaz:
I agree with @Keithbsmiley. Seems too heavy. Also, we haven't had much churn around adding new languages.

PR #353, opened 2015-02-27
Update tat script to accept optional session name

graysonwright:
If no session name is provided, it will generate one from the current directory.

### DIFF
--- a/bin/tat
+++ b/bin/tat
@@ -2,7 +2,8 @@
 #
 # Attach or create tmux session named the same as current directory.
 
-session_name="$(basename "$PWD" | tr . -)"
+path_name="$(basename "$PWD" | tr . -)"
+session_name=${1-$path_name}
 
 not_in_tmux() {
   [ -z "$TMUX" ]


bin/tat
```diff
@@ -2,7 +2,8 @@
 #
 # Attach or create tmux session named the same as current directory.
 
-session_name="$(basename "$PWD" | tr . -)"
+path_name="$(basename "$PWD" | tr . -)"
+session_name=${1-$path_name}
```
gabebw: I think you mean `#{1:-"#path_name"}`?

christoomey:
My only concern would be edge cases like spaces and dots in the provided name, but I guess that is on the user? Overall I like the change.

graysonwright:
yeah, I've had to remember to put dashes in the name when appropriate (e.g. `graysonwright-com`), but it hasn't really slowed me down.

I've been using this for a while now and I really like it. One workflow improvement I like: to start a new session, type `tat projectname`, and *then* `cd` to the directory. Otherwise, I'd have to `cd` to the new directory from the original session, which I feel pollutes my original session.

christoomey:
@graysonwright For the other workflow, I've been using there following mapping:

``` tmux
bind C-b send-keys 'tat && exit' 'C-m'
```

With that in your tmux.conf, you can split open a pane, cd to the desired directory (I rely on cdpath so I can typically do `cd projec<tab>` from anywhere and get to `~/code/apps/project-name`), then `C-b` in tmux will "break out" a new session rooted in that directory, cleaning up the pane from the original session. You can see it in action in the "Breaking Out Sessions" section of https://upcase.com/videos/tmux-advanced-workflow.

bin/tat
graysonwright: @gabebw I'm using this form in my dotfiles, and it's been working fine (I'm using ZSH).

what do the `:` and quotes add?

gabebw: Interesting! I assumed that `${x-"default"}` would print out the value of `$x`, then the literal string `-default`. It looks like this behaves exactly the same as `${x:-"default"}`, i.e. it falls back only if `$x` (or, in this case, `$1`), is not set.

I recommended double quotes because what if the path name has spaces? But it seems like it handles them sanely even without double quotes.

gabebw: In conclusion: it seems like my suggestion doesn't change the behavior. I'd still prefer `:-` to `-`, just because `:-` is more common, I think.

gylaz:
@graysonwright What's the status on this? Is it good to merge as is?

graysonwright:
@gylaz I've been using it with no complaints since I opened this PR back in February, and (as far as my workflow goes, it's my favorite part of `tat`). I'm good to merge it if others find it valuable.

@christoomey what do you think?

christoomey:
Essentially transparent to my use case, and seems useful, so I'm :+1: on it.

bin/tat
bernerdschaefer: @gabebw @graysonwright For future reference, `:-` uses the default when `$x` is either empty or undefined, while `-` only sets the default if undefined.

```
$ X=value eval 'echo ${X:-"default"}'
value
$ eval 'echo ${X:-"default"}'
default
$ X= eval 'echo ${X:-"default"}'
default

$ X=value eval 'echo ${X-"default"}'
value
$ eval 'echo ${X-"default"}'
default
$ X= eval 'echo ${X-"default"}'

```

graysonwright: ah, that's good to know. Thanks, Bernerd!

PR #354, opened 2015-02-28
Map <Leader>a in Vim to run all specs

ventsislaf:
Since we already have key mappings for `RunCurrentSpecFile`,
`RunNearestSpec` and `RunLastSpec`, I thought it would be good idea to
have for `RunAllSpecs` as well. It's using the keys suggested in
https://github.com/thoughtbot/vim-rspec


gylaz:
This has come up before in #312 (has my reasoning for why I closed it).

ventsislaf:
I haven't seen this. Thanks for sharing the link. It's fair enough.

PR #358, opened 2015-03-12
Let ctrl-p find files with a leading dot

derekprior:
I am constantly forgetting that I can't use `ctrl-p` to open
`.travis.yml` or any other file with a leading `.`. This change comes
about after some discussion in general on how to handle this. Passing
`--hidden` to the `ag` command allows it to find files with a leading
`.`.

Unfortunately, this also includes the content of your `.git`
directory. To overcome this, we add `/.git/` to `agignore`.

### DIFF
--- a/agignore
+++ b/agignore
@@ -1,3 +1,4 @@
 log
 tags
 tmp
+.git/
--- a/vimrc
+++ b/vimrc
@@ -68,7 +68,7 @@ if executable('ag')
   set grepprg=ag\ --nogroup\ --nocolor
 
   " Use ag in CtrlP for listing files. Lightning fast and respects .gitignore
-  let g:ctrlp_user_command = 'ag %s -l --nocolor -g ""'
+  let g:ctrlp_user_command = 'ag %s -l --nocolor --hidden -g ""'
 
   " ag is fast enough that CtrlP doesn't need to cache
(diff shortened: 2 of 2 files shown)


gabebw:
Looks fine to me.

JoelQ:
:+1:

iwz:
yessss

jferris:
Good call.

agignore
```diff
@@ -1,3 +1,4 @@
 log
 tags
 tmp
+/.git/
```
gylaz: Do we need the leading slash here? How about `.gitignore` files that may be in subdirectories?

gabebw: I don't think we should ignore `.gitignore` files - I may want to find those with ctrl-p. Plus, `.gitignore` files are not in `.git` directories, and so are correctly not ignored by this line.

graysonwright: I'm in favor of removing the leading slash - git repos can be nested with submodules, or I might want to run ctrl-p from a folder that contains multiple repos.

derekprior: removed the leading slaash, kept the trailing to make sure we're ignoring a directory and not a file named `.git`. Should probably update log and tmp similarly, but I'll do that in a follow-up

PR #359, opened 2015-03-12
Remove GitHub colorscheme

jessieay:
* Every time I pull updates from this repo into my own dotfiles, I have to
  reconfigure my preferred theme (solarized)
* Seems like there is not a majority or plurality of thoughbotters using a
  single colorscheme
* Easiest to just not specify a colorscheme
* See https://forum.upcase.com/t/why-is-the-default-vim-theme-on-dotfiles-is-github/4232

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -74,11 +74,6 @@ if executable('ag')
   let g:ctrlp_use_caching = 0
 endif
 
-" Color scheme
-colorscheme github
-highlight NonText guibg=#060606
-highlight Folded  guibg=#0A0A0A guifg=#9090D0
-
 " Make it obvious where 80 characters is
 set textwidth=80
 set colorcolumn=+1
--- a/README.md
+++ b/README.md
@@ -73,6 +73,13 @@ Your `~/.gitconfig.local` might look like this:
       name = Dan Croak
       email = dan@thoughtbot.com
 
+Your `~/.vimrc.local` might look like this:
+
+    " Color scheme
+    colorscheme github
+    highlight NonText guibg=#060606
+    highlight Folded  guibg=#0A0A0A guifg=#9090D0
+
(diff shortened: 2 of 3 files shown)


vimrc
```diff
@@ -75,7 +75,6 @@ if executable('ag')
 endif
 
 " Color scheme
-colorscheme github
 highlight NonText guibg=#060606
```
jferris: Should we also take out these `highlight` preferences?

jferris:
Good call! Should be easy to specify your own color scheme in `.local` files.

croaky:
I use it, enjoy it, and am just as happy to have it in my personal dotfiles rather than thoughtbot dotfiles. Tested with `.local` and works as expected. Looks good to merge.

vimrc
jessieay: I think this is for highlighting extra whitespace, which I think is good for any colorscheme. Reference https://wincent.com/blog/making-vim-highlight-suspicious-characters

Thoughts?

derekprior: I think it's a good idea, but we don't know what background people are using and can't be sure this will be an affective highlight.

gylaz: I didn't realize trailing whitespace was highlighted, likely due to my dark color-scheme.

vimrc
```diff
@@ -75,7 +75,6 @@ if executable('ag')
...
 highlight Folded  guibg=#0A0A0A guifg=#9090D0
```
gylaz: Should we remove this too?

vimrc
jessieay: yeah, same actually. Will remove.

jessieay:
@croaky added the local color setup to README - re-review?

croaky:
Looks good!

PR #362, opened 2015-03-20
Add git alias for hub pr without opening editor

danielnolan:
* Default behavior of hub when opening a pull-request with only one
  commit is to use the last commit title and body as the title and body
  for the pull-request, but hub opens the editor to allow you to make
  changes to the title and body. Since the commit message should be
  sufficient there is no need to have hub open the editor.
* This alias allows the last commit message to be used but
  prevents hub from opening the editor to make changes.
* Opening a pull-request like this was suggested by the hub developer in this issue https://github.com/github/hub/issues/722

### DIFF
--- a/gitconfig
+++ b/gitconfig
@@ -18,6 +18,7 @@
   delete-branch = !sh -c 'git push origin :refs/heads/$1 && git branch -D $1' -
   merge-branch = !git checkout master && git merge @{-1}
   pr = !hub pull-request
+  pr-noedit = !GIT_EDITOR=true hub pull-request
   rename-branch = !sh -c 'old=$(git current-branch) && git branch -m $old $1 && git push origin --set-upstream $1 && git push origin --delete $old' -
   st = status
   up = !git fetch origin && git rebase origin/master


gylaz:
I'm not sure I've felt the pain of having the editor open when I submit a PR. It serves as a nice "review" stage before submitting the PR to GitHub. In favor of keeping the list of aliases small -- and avoid this becoming a swiss utility knife -- I'm :-1: on this.

danielnolan:
@gylaz I see your point about it becoming a swiss army knife. Having the editor open IMO is just redundant. Based on this issue with hub github/hub#722 I'm not the only one that finds it to be. If I am in a feature branch all commits squashed, and I'm ready to push the branch up. It should be as easy as 
`g push && g pr-noedit` to push the branch and open the pull request. Why review the commit message I just wrote? If it doesn't work for everyone I understand, it will find a nice home in my local dotfiles.

croaky:
I've felt the same as @gylaz. I am used to `git commit -v` opening my editor for final review while crafting the commit message and it's nice that the pull request creation feels similar.

jferris:
I also like the review step before submitting a pull request. It's one last change to make sure I wrote a useful message before I ask other people to spend time on it.

PR #363, opened 2015-03-23
Remove 'github' colorscheme from README.md

korabh:
'github' colorscheme was deleted in b3cb238


croaky:
That section is showing an example of what could be in an individual's vimrc.local. The current example seems as good as any, perhaps better because it can help provide a backwards compatible path for anyone updating from an earlier version of these dotfiles.

PR #364, opened 2015-04-09
Ignore vendor/ during ag searches

gylaz:


jferris:
Makes sense to me to have `vendor` in there.

I think `coverage` should be in `gitignore` for any project which generates it, so it would already be ignored by ag. Does that sound right?

gylaz:
> I think coverage should be in gitignore for any project which generates it, so it would already be ignored by ag. Does that sound right?

Sounds correct. Just checked and suspenders has `coverage/` in `.gitignore`. Will update for that.

jferris:
Looks good to me!

andyw8:
@gylaz @jferris I'm curious why `vendor` was added here.

I've been working on a 'legacy' Rails app which has code in the `vendor/gems` and `vendor/engines` directories. I was initially confused as to why `ag` wasn't showing results for these directories.

I can understand that excluding `vendor/assets` would make sense since that generally third-party JS libraries, etc.

PR #365, opened 2015-04-10
Exclude `node_modules` and `bower_components`

seanpdoyle:
This is equivalent to excluding `vendor/assets` in a Rails project.

JS Dependency trees (especially npm packages) can be extremely noisy.

### DIFF
--- a/agignore
+++ b/agignore
@@ -3,3 +3,5 @@ tags
 tmp
 vendor
 .git/
+bower_components/
+node_modules/


BlakeWilliams:
:+1:

gylaz:
Is this something that typically gets checked in, or do these get gitignored?

BlakeWilliams:
I usually add it to .gitignore, but some projects don't.

seanpdoyle:
They're typically git-ignored, though my `ag` still includes them in searches unless I add these patterns to `.agignore`.

gylaz:
`ag` should respect entries in your `.gitignore` ([source](https://github.com/ggreer/the_silver_searcher#whats-so-great-about-ag)), if not there's a problem with your install.

How widely used is Bower still?

derekprior:
Are these required at runtime? I know that at least with node, without checking these dependencies in it's basically impossible to guarantee version consistency between deploys. npm has `shrinkwrap` that's meant to ease this some, but it's a manual process.

On our node project last year we tried everything avoid checking in `node_modules` but it boiled down to two options:

1. Check them in and get version consistency
2. Don't check them in and take your chances.

I'm not certain, but I thought Bower was the same in this vein.

derekprior:
Ah sorry, this is to add them to .agignore. Carry on (I wouldn't add these to gitignore to get the effect of agignore though).

BlakeWilliams:
This is .agignore

gylaz:
Sounds like these are directories that are often added to source control. :+1: on merging this.

seanpdoyle:
Any opposition to me merging this?

PR #356, opened 2015-03-01
Add git delete-branch completion

keith:
Closes https://github.com/thoughtbot/dotfiles/issues/355 via https://coderwall.com/p/wjuoag/creating-custom-git-commands-autocompletion

PR #367, opened 2015-04-12
Move Git aliases to subcommand scripts

croaky:
Any executable script on you PATH
that is named `git-some-name`
will be available as a git subcommand,
which means you could do `git some-name` to run the script.

Git adds them to `git help -a` under the title
"git commands available from elsewhere on your $PATH",
which will then power the auto completion,
so that will also work for any command you add.

http://blog.zamith.pt/blog/2014/11/05/supercharging-your-git/

Examples of other projects that structure their dotfiles like this:

https://github.com/robbyrussell/oh-my-zsh
https://github.com/tj/git-extras
https://github.com/holman/dotfiles

### DIFF
--- a/bin/git-up
+++ b/bin/git-up
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+set -e
+
+git fetch origin
+git rebase origin/master "$@"
--- a/gitconfig
+++ b/gitconfig
@@ -11,16 +11,7 @@
   ca = commit --amend -v
   ci = commit -v
   co = checkout
-  co-pr = !sh -c 'git fetch origin pull/$1/head:pr/$1 && git checkout pr/$1' -
-  create-branch = !sh -c 'git push origin HEAD:refs/heads/$1 && git fetch origin && git branch --track $1 origin/$1 && cd . && git checkout $1' -
-  ctags = "!sh -c '[ -f .git/hooks/ctags ] || git init; .git/hooks/ctags' git-ctags"
-  current-branch = !sh -c 'git rev-parse --abbrev-ref HEAD' -
-  delete-branch = !sh -c 'git push origin :refs/heads/$1 && git branch -D $1' -
-  merge-branch = !git checkout master && git merge @{-1}
-  pr = !hub pull-request
-  rename-branch = !sh -c 'old=$(git current-branch) && git branch -m $old $1 && git push origin --set-upstream $1 && git push origin --delete $old' -
   st = status
-  up = !git fetch origin && git rebase origin/master
 [core]
   excludesfile = ~/.gitignore
   autocrlf = input
--- a/bin/git-co-pr
+++ b/bin/git-co-pr
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+set -e
+
+git fetch origin "pull/$1/head:pr/$1"
+git checkout "pr/$1"
--- a/bin/git-create-branch
+++ b/bin/git-create-branch
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+set -e
+
+git push origin "HEAD:refs/heads/$1"
+git fetch origin
+git branch --track "$1" "origin/$1"
+cd .
+git checkout "$1"
--- a/bin/git-ctags
+++ b/bin/git-ctags
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+set -e
+
+[ -f .git/hooks/ctags ] || git init
+.git/hooks/ctags
--- a/bin/git-current-branch
+++ b/bin/git-current-branch
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+set -e
+
+git rev-parse --abbrev-ref HEAD
--- a/bin/git-delete-branch
+++ b/bin/git-delete-branch
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+set -e
+
+git push origin :refs/heads/"$1"
+git branch --delete "$1"
--- a/bin/git-merge-branch
+++ b/bin/git-merge-branch
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+set -e
+
+git checkout master
+git merge "@{-1}"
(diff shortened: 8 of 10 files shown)


teoljungberg:
:+1: I prefer having "pure git" alises in `~/.gitconfig` and the shell-influenced ones in my `$PATH`

gylaz:
:+1:

jferris:
This looks like an improvement to me.

gitconfig
```diff
@@ -11,16 +11,8 @@
   ca = commit --amend -v
   ci = commit -v
   co = checkout
-  co-pr = !sh -c 'git fetch origin pull/$1/head:pr/$1 && git checkout pr/$1' -
-  create-branch = !sh -c 'git push origin HEAD:refs/heads/$1 && git fetch origin && git branch --track $1 origin/$1 && cd . && git checkout $1' -
   ctags = "!sh -c '[ -f .git/hooks/ctags ] || git init; .git/hooks/ctags' git-ctags"
```
croaky: I'm a little confused about how to translate this one. What does that trailing `git-ctags` mean?

jferris: I'm not sure about that one, either.

@thoughtbot/shell ?

mike-burns: It's the command name, for e.g. ps(1) or arg[0] purposes.

mike-burns: For more, here is what I found in sh(1) on Debian (so, dash(1)):

```
     dash -c [-aCefnuvxIimqVEb] [+aCefnuvxIimqVEb] [-o option_name]
          [+o option_name] command_string [command_name [argument ...]]
...
           -c               Read commands from the command_string operand
                            instead of from the standard input.  Special
                            parameter 0 will be set from the command_name op‐
                            erand and the positional parameters ($1, $2, etc.)
                            set from the remaining argument operands.
```

jferris: Ah, so that means you could just move the contents of `-c` into `git-ctags` and it would work, because the process will actually be `git-ctags` at that point.

pbrisbin: Also, once it's invoked as an external script named `git-ctags`, that'll happen naturally.

gitconfig
```diff
@@ -11,16 +11,8 @@
...
-  current-branch = !sh -c 'git rev-parse --abbrev-ref HEAD' -
-  delete-branch = !sh -c 'git push origin :refs/heads/$1 && git branch -D $1' -
-  merge-branch = !git checkout master && git merge @{-1}
-  pr = !hub pull-request
```
calebthompson: This one actually could have just been `pr = pull-request`, provided hub is aliased in the dotfiles?

calebthompson:
I'm all for this. One thing I noticed is that you replaced a lot of `this && that` with

```
this
that
```

Which is more closely equivalent to `this; that`. We just lose the short-circuit failures in this case.

I think the best solution is just to keep

```
this && that
```

as one-line scripts, but longer ones could be split over multiple lines.

You could also use `set -e`.

gitconfig
```diff
@@ -11,16 +11,8 @@
   ca = commit --amend -v
   ci = commit -v
   co = checkout
-  co-pr = !sh -c 'git fetch origin pull/$1/head:pr/$1 && git checkout pr/$1' -
-  create-branch = !sh -c 'git push origin HEAD:refs/heads/$1 && git fetch origin && git branch --track $1 origin/$1 && cd . && git checkout $1' -
   ctags = "!sh -c '[ -f .git/hooks/ctags ] || git init; .git/hooks/ctags' git-ctags"
-  current-branch = !sh -c 'git rev-parse --abbrev-ref HEAD' -
```
calebthompson: I believe this can be `current-branch = rev-parse --abrev-ref HEAD`? That's what is in my dotfiles. I don't know where the trailing hyphen comes from.

bin/git-up
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+git fetch origin
+git rebase origin/master
```
calebthompson: I could be wrong, but `pull` is functionally equivalent to `fetch && merge`, so isn't this equivalent to `git pull --rebase origin master`?

calebthompson:
```
       -r, --rebase[=false|true|preserve]
           When true, rebase the current branch on top of the upstream branch after fetching. If there is a remote-tracking branch corresponding to the upstream branch and the upstream branch was rebased
           since last fetched, the rebase uses that information to avoid rebasing non-local changes.
```

calebthompson:
> which will then power the auto completion,

I get completion for aliases. Do you not?

gitconfig
jferris: It's [not aliased](https://github.com/thoughtbot/dotfiles/blob/master/aliases). I personally prefer to maintain the distinction between `git` and `hub`.

jferris:
> You could also use `set -e`.

I'd be in favor of this.

calebthompson:
While I like the idea, its benefit is that it moves shell into shell files, with highlighting and such, I don't think that it gains us anything in the way of functionality. I'd never even used `git help -a` and don't think that it would benefit me more than `git somte[tab]` does.

Which is just to say, I'd like the commit message and PR message to reflect that.

croaky:
> I get completion for aliases. Do you not?

The aliases themselves autocomplete but their arguments do not. My hope was that this change could make https://github.com/thoughtbot/dotfiles/pull/356 easier for `git delete-branch dc-<tab>` but it hasn't appeared to help.

gitconfig
croaky: Cool, thanks 9ce8388

croaky:
> You could also use `set -e`.

Good catch. I did that in 340a3b6

calebthompson:
Ah, I see. You'll need to add some compdefs to zsh completion then for these custom commands. You can effectively "alias" completions so they function like another command or add your own. In the case of git-delete-branch case you might want only local branches for example.

croaky:
Good to merge?

jferris:
Looks good to me.

gylaz:
Seems I've lost tab completion after this. `git delete-br<TAB>` does not auto complete for me.

Jbigdaddylewis23:
> > You could also use `set -e`.
> 
> I'd be in favor of this.
>

PR #368, opened 2015-04-15
Use vim-plug over vundle

geoffharcourt:
*Note: I accidentally deleted the branch associated with https://github.com/thoughtbot/dotfiles/pull/345 and Github won't allow me to re-associate the branch with that PR, so I'm opening a new PR. Apologies for the convoluted paper trail.*

[vim-plug](https://github.com/junegunn/vim-plug) has a number of
advantages over Vundle:

* Installs and updates plugins very quickly in parallel
* Can lock plugins at versions/tags
* Can rollbacks updates (useful if a plugin breaks) and take/reload
  snapshots of current state
* Optionally lazily-load plugins when their relevant command is invoked
* Execute post-update hooks for plugins with compiled extensions, etc.

vim-plug uses a DSL very close to Vundle (simplest form is `Plug` vs.
`Plugin`), and here it is set to continue to use the same plugin
location that Vundle was using before.

After updating, users will need to
1. Rename `Plugin` lines in `.vimrc.bundles.local` to use `Plug`
2. Run `:PlugInstall` (the post-up hook does this)

### DIFF
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -2,7 +2,8 @@
 
 touch $HOME/.psqlrc.local
 
-if [ ! -e $HOME/.vim/bundle/Vundle.vim ]; then
-  git clone https://github.com/gmarik/Vundle.vim.git $HOME/.vim/bundle/Vundle.vim
+if [ ! -e $HOME/.vim/autoload/plug.vim ]; then
+  curl -fLo $HOME/.vim/autoload/plug.vim --create-dirs \
+      https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 fi
-vim -u $HOME/.vimrc.bundles +PluginInstall +PluginClean! +qa
+vim -u $HOME/.vimrc.bundles +PlugInstall +PlugClean! +qa
--- a/vimrc.bundles
+++ b/vimrc.bundles
@@ -2,36 +2,51 @@ if &compatible
   set nocompatible
 end
 
-filetype off
-set rtp+=~/.vim/bundle/Vundle.vim/
-call vundle#begin()
+" Shim command and function to allow migration from Vundle to vim-plug.
+function! PluginToPlug(arg, ...)
+  echom "You are using Vundle's `Plugin` command to declare plugins. Dotfiles now uses vim-plug for plugin mangagement. Please rename uses of `Plugin` to `Plug`. Plugin was '".a:arg."'."
+  let vim_plug_options = {}
 
-" Let Vundle manage Vundle
-Plugin 'gmarik/Vundle.vim'
+  if a:0 > 0
+    if has_key(a:1, 'name')
+      let name = a:1.name
+      let vim_plug_options.dir = "$HOME/.vim/bundle/".a:1.name
+    endif
+
+    if has_key(a:1, 'rtp')
+      let vim_plug_options.rtp = a:1.rtp
+    endif
+  endif
+
+  Plug a:arg, vim_plug_options
+endfunction
+
+com! -nargs=+  -bar Plugin call PluginToPlug(<args>)
+
+call plug#begin('~/.vim/bundle')
 
 " Define bundles via Github repos
-Plugin 'christoomey/vim-run-interactive'
-Plugin 'kchmck/vim-coffee-script'
-Plugin 'ctrlpvim/ctrlp.vim'
-Plugin 'pbrisbin/vim-mkdir'
-Plugin 'scrooloose/syntastic'
-Plugin 'slim-template/vim-slim'
-Plugin 'thoughtbot/vim-rspec'
-Plugin 'tpope/vim-bundler'
-Plugin 'tpope/vim-endwise'
-Plugin 'tpope/vim-eunuch'
-Plugin 'tpope/vim-fugitive'
-Plugin 'tpope/vim-rails'
-Plugin 'tpope/vim-repeat'
-Plugin 'tpope/vim-surround'
-Plugin 'vim-ruby/vim-ruby'
-Plugin 'vim-scripts/ctags.vim'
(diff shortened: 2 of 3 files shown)


vimrc.bundles
```diff
@@ -2,36 +2,51 @@ if &compatible
...
+  echom "You are using Vundle's `Plugin` command to declare plugins. Dotfiles now uses vim-plug for plugin mangagement. Please rename uses of `Plugin` to `Plug`. Plugin was '".a:arg."'."
+  let vim_plug_options = {}
 
-" Let Vundle manage Vundle
-Plugin 'gmarik/Vundle.vim'
+  if a:0 > 0
+    if has_key(a:1, 'name')
+      let name = a:1.name
+      let vim_plug_options.dir = "~/.vim/bundle/".a:1.name
+    endif
```
croaky: Any advantage to using $HOME instead of ~?

hooks/post-up
```diff
@@ -2,7 +2,8 @@
 
 touch $HOME/.psqlrc.local
 
-if [ ! -e $HOME/.vim/bundle/Vundle.vim ]; then
-  git clone https://github.com/gmarik/Vundle.vim.git $HOME/.vim/bundle/Vundle.vim
+if [ ! -e $HOME/.vim/autoload/plug.vim ]; then
+  curl -fLo ~/.vim/autoload/plug.vim --create-dirs \
```
croaky: ~ to $HOME?

vimrc.bundles
geoffharcourt: I think there is. Let me update that.

croaky:
That migration path / notice / backwards compatibility shim is cool.

This seems like a good idea to me.

geoffharcourt:
@croaky I'm leaving the last update un-squashed for now, but if there's consensus approval, I'll get it into one commit for merging.

vimrc.bundles
```diff
@@ -2,36 +2,51 @@ if &compatible
...
+
+    if has_key(a:1, 'rtp')
+      let vim_plug_options.rtp = a:1.rtp
+    endif
+  endif
+
+  Plug a:arg, vim_plug_options
+endfunction
+
+com! -nargs=+  -bar Plugin call PluginToPlug(<args>)
```
gylaz: Any thoughts on how long we'd need to keep this around?

vimrc.bundles
```diff
@@ -2,36 +2,51 @@ if &compatible
...
+
+call plug#begin('~/.vim/bundle')
```
gylaz: Should this be `call plug#begin('~/.vim/plugged')`? Or are we keeping `bundle` for backwards compatibility?

gylaz:
This is great @geoffharcourt!

vimrc.bundles
geoffharcourt: I think three months or so is probably fine. I pull down changes fairly consistently, but I'm not sure if others might pull changes far less frequently, in which case maybe a longer time is fine. I could add a comment to mark the expiration date for when we'd want to delete it if that's useful.

vimrc.bundles
geoffharcourt: I don't think we should move the plugins around. Moving them might introduce a hitch into the transition, and there's no big benefit in having them in a new folder. I think a Ruby developer who is new to Vim might also find the folder `bundle` to be more familiar if they are starting to dig through dotfiles.

Also, pathogen uses `bundle`, so that seems to be a more general practice than just Vundle.

croaky:
Good to merge, @gylaz?

git_template/hooks/pre-commit
```diff
@@ -1,7 +1,46 @@
...
+new_names=`$git_diff HEAD` || exit
+case "
+$new_names" in
+  */-* | *'
+'-*)
+    echo "File name component begins with '-'."
+    exit 1;;
+esac
+
+exec git diff-index --check --cached HEAD --
```
gylaz: Was this accidentally added?

geoffharcourt: Ugh, I was fooling around with Spacebox and that got included. Fixing it now.

geoffharcourt:
I fixed the extra file @gylaz spotted in review, and squashed the commits down after rebasing with today's merge.

gylaz:
Thanks again @geoffharcourt!

geoffharcourt:
Woohoo! Thanks, excited to see this in upstream.

seanpdoyle:
What additional steps are necessary once we've `:%s/Plugin/Plug/g`?

I'm failing:

```
You are using Vundle's `Plugin` command to declare plugins. Dotfiles now uses vim-plug for plugin mangagement. Please rename uses of `Plugin` to `Plug`. Plugin was 'tpope/vim-unimpaired'.
E169: Command too recursive
Error detected while processing /Users/Sean/.vimrc.bundles:
line   52:
E117: Unknown function: plug#end
Error detected while processing /Users/Sean/.vimrc.local:
```

geoffharcourt:
@seanpdoyle, can you paste your current .vimrc.local (I'm assuming your changes aren't checked in yet). 

Does vim load for you with just the warnings before you changed `Plugin` -> `Plug`?

seanpdoyle:
@geoffharcourt https://github.com/seanpdoyle/dotfiles.local/blob/8e5bdffae159cd4ef04bbe22207bc824c9ba22dc/vimrc.bundles.local changes have been pushed.

I get the same warning no matter what

seanpdoyle:
@geoffharcourt never realized there was a `vim-plug` installation step: https://github.com/junegunn/vim-plug#usage

I'll try that

geoffharcourt:
@seanpdoyle, that's part of the post-up hook after you run `rcup`. I've gotten burned by changes there before after forgetting to to run it certain dotfiles updates.

PR #369, opened 2015-04-17
Ignore `.env.local` instead of `.env`.

lpender:
Here is an common occurrence that the current `.gitignore` supports:

1. The `.sample.env` (or like) file is updated with environment variables after a feature is merged.
2. A developer pulls the latest from master and runs the test suite.
3. It fails with an opaque error message.

A more robust workflow versions keys used in local development in the `.env` file and overrides these locally with the `.env.local` file.

See https://github.com/bkeepers/dotenv#should-i-commit-my-env-file

### DIFF
--- a/gitignore
+++ b/gitignore
@@ -1,7 +1,7 @@
 .DS_Store
 *.sw[nop]
 .bundle
-.env
+.env.local
 db/*.sqlite3
 log/*.log
 rerun.txt


jferris:
I think this is probably a good approach, but I also think we should try it out on a project before committing it to our dotfiles.

gylaz:
This seems like a better candidate for `.gitignore` at the project level.

lpender:
Perhaps this PR belongs in the `Suspenders` repo...

lpender:
https://github.com/thoughtbot/dotfiles/pull/369

lpender:
https://github.com/thoughtbot/suspenders/pull/563

tute:
Hey @lpender! Do you want to reopen this PR, in light of https://github.com/thoughtbot/suspenders/pull/563#discussion_r39910388? Thank you!

lpender:
Hi @tute! Sorry I wasn't receiving GH notifications for some reason. Taking a look now.

lpender:
Ok I'm going to reopen--I believe that was the conclusion of the other conversation, yes?

tute:
Already discussing this change in https://github.com/thoughtbot/dotfiles/pull/442. Thanks, Lee!

PR #371, opened 2015-04-21
Added git_template/info/exclude for Xcode

Tambling:
Xcode >=6.0.1 throws an error when creating a new git-tracked project without an 'exclude' file. Created the exclude file to correspond to the existing gitignore file.

### DIFF
--- a/git_template/info/exclude
+++ b/git_template/info/exclude
@@ -1,0 +1,6 @@
+# git ls-files --others --exclude-from=.git/info/exclude
+# Lines that start with '#' are comments.
+# For a project mostly in C, the following would be a good set of
+# exclude patterns (uncomment them if you want to use them):
+# *.[oa]
+# *~


gylaz:
Can anyone from @thoughtbot/ios pitch in on this?

gfontenot:
That doesn't sound right to me. I've never created that file, and I've never seen any issues.

jkrmr:
I was seeing that issue. This resolved it for me.

regnerjr:
I have this problem as well. Have not tried this fix.

jakecraige:
Rather than adding all the gitignore to the exclude file, what about adding the default git exclude file? There's no need to duplicate the .gitignore into the exclude file since they work together

It looks like this:

```
# git ls-files --others --exclude-from=.git/info/exclude
# Lines that start with '#' are comments.
# For a project mostly in C, the following would be a good set of
# exclude patterns (uncomment them if you want to use them):
# *.[oa]
# *~
```

On my machine, can be found at `/usr/local/share/git-core/templates/info/exclude`

Tambling:
@jakecraige You're absolutely right. I'm sure I thought I had a reason to duplicate the .gitignore, but in retrospect, it was probably my enthusiasm getting the better of me. I've amended my original commit to reflect your suggestion.

jakecraige:
Great!

I'm :+1: unless anyone else has an objections to this.

gylaz:
Thanks @Tambling. Merged as 66491639e4e52538658d950e61bae85c7bd8c062.

Tambling:
^_^ Thanks @jakecraige & @gylaz!

PR #372, opened 2015-04-24
Fix rbenv warning on terminal startup

iamarmanjon:
It seems rbenv has an error message, I did the fix stated here: https://github.com/sstephenson/rbenv/issues/487.

I don't if anyone is having problem for that too. It is on my end.


gylaz:
I'm not getting it on version `0.4.0` of `rbenv`, and have not heard any reports from others. Closing this.

PR #376, opened 2015-05-02
Update shim that helps migrate users from Vundle to vim-plug to catch bo...

joshleichtung:
...th Plugin and Bundle commands for users that are still using Vundle's older syntax.


gylaz:
Thanks @joshleichtung! Merged as ae4fa0a19ac747d4ab7e2316d68c556276fb82a8.

PR #379, opened 2015-05-06
Remove matchit.zip plugin

danielnolan:
* This plugin is no longer maintained and is included in vim by default.
* Per the docs at https://github.com/vim-scripts/matchit.zip
  Since vim 6.0, matchit.vim has been included in the standard vim distribution,
  under the macros/ directory; the version here may be more recent.
* The version included in the vim macros directory is the same version
  being added in the vimrc.bundles file from github version 1.13.2

### DIFF
--- a/vimrc.bundles
+++ b/vimrc.bundles
@@ -42,7 +42,6 @@ Plug 'tpope/vim-repeat'
 Plug 'tpope/vim-surround'
 Plug 'vim-ruby/vim-ruby'
 Plug 'vim-scripts/ctags.vim'
-Plug 'vim-scripts/matchit.zip'
 Plug 'vim-scripts/tComment'
 
 if filereadable(expand("~/.vimrc.bundles.local"))


derekprior:
:+1:

JoelQ:
Seems reasonable to me :+1:

gylaz:
Merged as 58bb228f7531a69d4c656fd14ac3712250c5358a. Thanks @danielnolan!

tsigo:
Unfortunately matchit.zip is no longer loaded anywhere, so the `%` mapping to jump, for example, between the start and end of a Ruby block, no longer works. I think we need to add `source $VIMRUNTIME/macros/matchit.vim` to vimrc now to get this working again.

geoffharcourt:
@tsigo, what version of Vim are you using? With both 7.4.712 and Neovim I'm getting the % jumping behavior without any explicit call to Matchit or the plugin being loaded.

tsigo:
@geoffharcourt Odd. Running 7.4.712 as well, and didn't see this behavior until I added that source line to vimrc.

geoffharcourt:
I think you must have a plugin that's hijacking the behavior of `%`. When I run vim 7.4.712 with no .vimrc at all, I get the jumping behavior with `(` and `)` in text files.

tsigo:
@geoffharcourt Sorry for the confusion, I'm talking about the behavior of `%` jumping between things like `if`, `elsif`, `end` and `do`/`end` in Ruby. This comes from matchit.zip.

geoffharcourt:
Ah, OK. I'm seeing the same thing you are now.

PR #380, opened 2015-05-08
Remove uncommon aliases

gylaz:


tute:
I use only one of those. I'm ok with moving it to my custom dotfile.

mcmire:
I've never used any of the ones you removed or variations thereof, so this makes sense to me.

croaky:
I use `m` and `s` but not the others.

jessieay:
I don't use any of these. Always feel like I should study / use the dotfile aliases more, but perhaps the fact that I've gone this long without using them means I don't need them at all...

andyw8:
I use `rk` and `s` almost every day,

mike-burns:
The `ln` one simply improves the output of ln(1), making it more clear what it just did - handy because the order of the arguments are confusing to some. That seems like something that should stay, unless we expect people to use ln(1) so often that they feel comfortable with it.

geoffharcourt:
I used to use the `rk` alias often, but I found there were enough scenarios where I had to run it with `bundle exec` that I gave up on it.

JoelQ:
I don't use any of these

derekprior:
I use some of these. I don't mind moving them to my own aliases, but I don't understand the line drawn here. To me there are a couple of different types of aliases define here:

**Shortcuts:**

* `b`
* `rk`
* `tlf*
* `e`
* `v`
* ...

** Best Practice/Common Error Avoidance:

* `ln`
* `mkdir`
* `migrate`
* `path`

I think everything in the latter category is worth keeping, but am not sure how we draw the line on the first category. I'd consider axing the entire first category.

jferris:
> I think everything in the latter category is worth keeping, but am not sure how we draw the line on the first category. I'd consider axing the entire first category.

I feel the same way about this.

gylaz:
> am not sure how we draw the line

My intention for this PR is to tease out that line.

I agree that best practices should stay.

MattMSumner:
I don't use any of these. :+1:

PR #385, opened 2015-05-13
Add "^K" keybind

jacquerie:
I'm used to running `^A^K` to delete the current line no matter where I am, but this isn't working using these dotfiles (instead, a `^K` is prepended to the line).

So far I've added this keybinding to `zshrc.local`, but maybe somebody else will find it useful.

NB: the sequence `^E^U` accomplishes the same thing and it's currently working.

### DIFF
--- a/zshrc
+++ b/zshrc
@@ -51,6 +51,7 @@ bindkey jj vi-cmd-mode
 # handy keybindings
 bindkey "^A" beginning-of-line
 bindkey "^E" end-of-line
+bindkey "^K" kill-line
 bindkey "^R" history-incremental-search-backward
 bindkey "^P" history-search-backward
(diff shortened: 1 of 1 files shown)


mike-burns:
LGTM.

jjlangholtz:
Would it also be worth adding the ```^B``` and ```^F``` bindings? These are pretty standard and I find it is useful to have the same key bindings here as well.

```
bindkey "^B" backward-char
bindkey "^F" forward-char
```

mike-burns:
@jjlangholtz I would welcome such an addition. Can you make a pull request?

gylaz:
Thanks @jacquerie. Merged as 13b182f973512f47a7a70ee7ea8f4d2afe28dfe5.

PR #386, opened 2015-05-19
Use C-s as tmux prefix

tute:
C-a allow use to trigger tmux prefix more easily than C-b, at the cost
of overriding the readline beginning of line. This change remaps it to
C-s, which has the same effect without that drawback.

### DIFF
--- a/tmux.conf
+++ b/tmux.conf
@@ -10,9 +10,7 @@ bind l select-pane -R
 bind-key -r C-h select-window -t :-
 bind-key -r C-l select-window -t :+
 
-# act like GNU screen
-unbind C-b
-set -g prefix C-a
+set -g prefix2 C-s
 
 # start window numbers at 1 to match keyboard order with tmux window order
 set -g base-index 1
@@ -32,8 +30,10 @@ set -g status-right ''
 # increase scrollback lines
 set -g history-limit 10000
 
-# switch to last pane
-bind-key C-a last-pane
+# prefix -> back-one-character
+bind-key C-b send-prefix
+# prefix-2 -> forward-incremental-history-search
+bind-key C-s send-prefix -2
 
 # Local config
 if-shell "[ -f ~/.tmux.conf.local ]" 'source ~/.tmux.conf.local'


croaky:
What does "overriding the readline beginning of line" mean or look like? Why is it a bad thing?

calleerlandsson:
Typing C-a in Readline moves the cursor to the beginning of the line. By using C-a as the tmux prefix, this Readline functionality will be unavailable. In GNU Screen, which prefix defaults to C-a, it is solved by mapping the Readline beginning of line function to C-a a.

jferris:
It seems like bad form to me to override the tmux prefix at all.

christoomey:
I'm a big :+1: on this (have only ever run with `C-s` as my prefix). 

`C-s` is either of "reverse-incremental-history-search", or pause flow control (essentially lock up terminal until you type `C-q`) which I find little to no use for. `C-a` is beginning of line in [readline keybindings][] which is a common shortcut, even in GUI applications on OS X (not sure about other environments) that I use constantly. My understanding is that the `C-a` prefix many use is a nod to GNU Screen which feels like a weaker reason to choose.

Incidentally, [our tmux course on Upcase](https://upcase.com/tmux) uses `C-s` since I was unable to record a video without messing up many (many) times using `C-a`.

@jferris are you recommending a return to the default of `C-b` for tmux?

Also of note, [we bind ctrl-a to beginning-of-line][] currently, but this is hidden by the tmux prefix.

[we bind ctrl-a to beginning-of-line]: https://github.com/thoughtbot/dotfiles/blob/d68ba146697bb8653c822a11227ba09eadd2c305/zshrc#L52



[readline keybindings]: http://www.bigsmoke.us/readline/shortcuts

jferris:
> Are you recommending a return to the default of C-b for tmux?

That makes the most sense to me. It's clear not everybody is going to agree on the same prefix, and if we're going to have a default, it probably makes sense to leave it as...the default.

What I'd really love is to find a way to let people have an alternate prefix while leaving C-b in place. It's always tricky when you jump onto somebody else's machine to pair and they have a different prefix. I've paired with people who use `b`, `a`, `s`, and more.

derekprior:
I'd be willing to do this. I've been wanting to for some time, just dreading the muscle memory hit.

> That makes the most sense to me. It's clear not everybody is going to agree on the same prefix, and if we're going to have a default, it probably makes sense to leave it as...the default.

The same argument could be made for `leader` in vim, but we change that as well.

I think the best argument against using the default is that tmux is nearly unusable without knowing the prefix combination whereas the leader key is a "nice to have" in vim.  For that reason maybe it would be better to leave it as the default and have people override in their local? Will tmux be okay with overriding prefix after binding a bunch of keys?

christoomey:
I just tested and tmux is fine to set the prefix after defining key bindings (unlike Vim which inlines the current `<leader>` value when mapping a binding).

With this in mind, I think it would be fine to move this to a `.local` configuration, and keep the default in our dotfiles.

image

Turns out you can have 2 prefix keys in tmux, allowing for @jferris's goal of having the default still work, while allow a user preference as well.

``` tmux
# default of C-b is the <prefix>, while C-s will also work as a prefix key
set -g prefix2 C-s
```

jferris:
> The same argument could be made for leader in vim, but we change that as well.

For what it's worth, I'd be fine leaving the `leader` key alone in Vim as well.

> I think the best argument against changing the default is that tmux is nearly unusable without knowing the prefix combination whereas the leader key is a "nice to have" in vim.

Yeah, I totally agree with this. When I pair with somebody, I generally assume all shell aliases and leader keys are unavailable. You can't really do that in tmux, so every time I pair, I have to go, "what's your tmux prefix?"

geoffharcourt:
There's a corresponding change [here](https://github.com/thoughtbot/dotfiles/blob/master/tmux.conf#L36) that should also be made so that "double-prefix" continues to switch to the last pane in tmux in the same way that "double-leader" jumps to the last buffer in vim.

tute:
> > The same argument could be made for leader in vim, but we change that as well.
> 
> For what it's worth, I'd be fine leaving the leader key alone in Vim as well.

I'm +1 on this too. I'll remove our tmux beginning-of-line mapping, and define `C-s` as prefix2.

tute:
I removed `unbind C-b`, and changed my line to `set -g prefix2 C-s`. When I try to run `C-s` works, but the default `C-b` prints `^B` in my zsh session, not sure how to fix that?

Also, to @geoffharcourt's comment on last pane, I'd like to be able to tell tmux that when any prefix key is repeated we should trigger that behavior, but it seems we need to specify both? That's the solution I did now, I wish there was a placeholder for "prefix".

geoffharcourt:
@tute, if we're going to support both `<C-b>` and `<C-s>` as prefixes, then I retract my request to move that mapping over. Using up two prefix mappings for something that for most people is likely used less often than the pane navigation from vim-tmux-navigator is probably overkill.

tmux.conf
```diff
@@ -33,7 +31,8 @@ set -g status-right ''
 set -g history-limit 10000
 
 # switch to last pane
-bind-key C-a last-pane
+bind-key C-b last-pane
+bind-key C-s last-pane
```
tute: @christoomey do you know if there's a `prefix` type of command so we don't need to specify both `C-b` and`C-s` here?

christoomey: @tute I'm not aware of `prefix` standin.

With that said, I'm wondering if we might want to take a different approach with these keys. `last-pane` is already bound to `;` by default which is comfortably on the home row, and if using vim-tmux-navigator also bound to (unprefixed) `C-\`.

As an alternative, I typically map the prefix to `send-prefix` such that double pressing the prefix will send those keystrokes through. `C-b` for instance is `back-one-character` which can be useful, and `C-s` is `forward-incremental-history-search`. Without the `send-prefix` mapping you essentially have no way to send these keys.

If interested, the configuration would be:

``` tmux
bind-key C-b send-prefix
bind-key C-s send-prefix -2
```

tute:
Updated PR, thanks for your feedback!

geoffharcourt:
@tute, this is great. That's an easy-enough mapping for folks to add on their own.

gylaz:
Looks good to me.

geoffharcourt:
@tute, I have noticed one annoying problem with `C-b` set as the prefix, which is that in selection mode for copying to the clipboard, `C-b` is no longer available to scroll backwards through the terminal to eyeball what you wanted to copy. Are thoughtboters using a different strategy when looking through a longer terminal output to find what they want to copy?

christoomey:
@geoffharcourt This won't solve `C-b` for you, but an alternative is `C-u` and `C-d` for half page jumps. I've always used them in Vim, pagers, etc. as I find a full page to be to jarring.

tute:
I do use `C-b` too; but I'll move to `C-u` (or remap my locals) in favor of default prefix key.

PR #373, opened 2015-04-30
Enable git completion of subcommands

gylaz:

PR #387, opened 2015-05-22
Add third-party site-functions to $fpath

gylaz:
Third-party completions get added to
`/usr/local/share/zsh/site-functions`.

Standard `$fpath` contains `/usr/share/zsh/site-functions`
(not the missing `local`).

By adding this to the `$fpath` git subcommands get completed correctly.
Related to https://github.com/thoughtbot/dotfiles/pull/373.

### DIFF
--- a/zshrc
+++ b/zshrc
@@ -9,7 +9,7 @@ setopt promptsubst
 export PS1='${SSH_CONNECTION+"%{$fg_bold[green]%}%n@%m:"}%{$fg_bold[blue]%}%c%{$reset_color%}$(git_prompt_info) %# '
 
 # load our own completion functions
-fpath=(~/.zsh/completion $fpath)
+fpath=(~/.zsh/completion /usr/local/share/zsh/site-functions $fpath)
 
 # completion
 autoload -U compinit


gylaz:
pinging @thoughtbot/shell

gabebw:
Perhaps a better commit message is: "Add third-party git completions"? Is that an accurate summary?

gylaz:
@gabebw I'm not certain that `/usr/local/share/zsh/site-functions` is used only for completions, and it's definitely not only for git.

gylaz:
The solution was found via: https://github.com/github/hub/issues/231#issuecomment-20879594

gylaz:
It might be the case that `zsh` installed via Homebrew already does this, but `zsh` that comes with OSX out fo the box does not.

MattMSumner:
This seems reasonable to me.

croaky:
For me (installed zsh via Homebrew) before:

```
% echo $fpath
/Users/croaky/.zsh/completion /usr/local/share/zsh/site-functions /usr/local/Cellar/zsh/5.0.7/share/zsh/functions
```

After:

```
% echo $fpath
/Users/croaky/.zsh/completion /usr/local/share/zsh/site-functions /usr/local/Cellar/zsh/5.0.7/share/zsh/functions
```

So, didn't hurt my environment, which tab-completed before and after, and should help others without zsh via Homebrew.

PR #388, opened 2015-05-23
Save file before running tests via vim-rspec

andyw8:
* Reduces the number of keystrokes to run a test
* Prevents the mistake of forgetting to save the spec file after editing it


gylaz:
Files already should be saved via https://github.com/thoughtbot/dotfiles/blob/master/vimrc#L13

andyw8:
Yep, my mistake! I had this in my `.vimrc.local` since I switched to using thoughtbot's files, so now I know it was never needed.

PR #390, opened 2015-06-01
Add config file for ctags

gylaz:
Allows to ignore certain directories from being ctag'd.

### DIFF
--- a/ctags
+++ b/ctags
@@ -1,0 +1,4 @@
+--recurse=yes
+--exclude=bower_components
+--exclude=node_modules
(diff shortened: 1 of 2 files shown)


jferris:
Does this need to be `~/ctags` or can we have them in project directories? I'm wondering if this would make more sense in Suspenders, since each project will have different vendor directories to ignore.

gylaz:
This can be on per project basis. I opened a PR is suspenders: https://github.com/thoughtbot/suspenders/pull/595

PR #391, opened 2015-06-02
git: Ignore `.agignore` in `.gitignore`

reshleman:
This excludes project-level `.agignore` files from git.

### DIFF
--- a/gitignore
+++ b/gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 *.sw[nop]
+.agignore
 .bundle
 .env
 db/*.sqlite3


JoelQ:
Can you explain your rationale behind this change? It's always great to have the "why" behind the change in the commit message.

reshleman:
@JoelQ Sure. Sorry for not including more detail originally.

Most (~95%?) of the time, files / directories that I want to exclude from `ag` searches are also files that should be excluded from source control altogether. Since `ag` ignores files / directories in `.gitignore`, I add those exclusions to `.gitignore` and commit it, so those exclusions are shared with other developers on the project.

On the other hand, I occasionally find that there are files / directories that I want to exclude from `ag` searches for project, but that should remain in source control. Most of the time, these are files that I consider noise in my searches, but other developers / designers may want to search (i.e., SVG images or even stylesheets). When I create a project-level `.agignore` for these files, I then have to add `.agignore` to the project's `.gitignore`.

When I do create an `.agignore` file for a project, I almost always also add it to that project's `.gitignore`. Adding `.agignore` to `~/.gitignore` excludes these user-level `ag` configurations from source control by default (but still allows them to be added manually, if necessary, with `git add -f .agignore`).

Not a *huge* pain point, to be sure, but it seems like a reasonable default. Happy to hear any feedback.

jferris:
It strikes me as strange that different developers would want different results in search. I feel like I'd actually want to commit the `.agignore` for the team. I've added (and committed) `.agignore` files to a couple projects, but like you said, it's usually covered by the `.gitignore`, so it doesn't come up often.

gylaz:
This seems like a user preference to me, and should not be a default. Something that should go into `~/.agignore.local` (when that is a possibility).

hiattp:
@gylaz +1 for `~/.agignore.local` being possible

PR #392, opened 2015-06-05
Add `ctags` support for various Ruby DSLs

teoljungberg:
I gave a lightning talk today about `ctags` and how I use it and showed
of some niceties I have configured. I thought I could share it with the
rest of the users of our dotfiles.

This adds support for:
- `attr_{reader,writer,accessor)`
- RSpec spec dsl:
  - context
  - describe
  - feature
  - it
  - scenario
  - shared_examples

### DIFF
--- a/ctags
+++ b/ctags
@@ -1,0 +1,2 @@
+--regex-ruby=/(^|[:;])[ \t]*([A-Z][[:alnum:]_]+) *=/\2/c,class,constant/
+--regex-ruby=/^[ \t]*attr_(reader|writer|accessor) (:[a-z0-9_]+, )*:([a-z0-9_]+)/\3/A,attr/


georgebrock:
This is probably relevant: https://trello.com/c/liLHC19x/14-improving-tags-in-ctags-files

gylaz:
Let's see how https://github.com/thoughtbot/suspenders/pull/595 turns out. `suspenders` might be a better place for this.

ctags
```diff
@@ -0,0 +1,8 @@
+--regex-ruby=/(^|[:;])[ \t]*([A-Z][[:alnum:]_]+) *=/\2/c,class,constant/
+--regex-ruby=/^[ \t]*attr_(reader|writer|accessor) (:[a-z0-9_]+, )*:([a-z0-9_]+)/\3/A,attr/
+--regex-ruby=/^[ \t]*context ['"](.*)['"].* do/\1/C,rspec context/
```
gabebw: So, the `attr_reader` line lets me `Ctrl-]` from `frump` to `att_reader :frump`, right?

But what does context/describe/feature/it/scenario let me do? Where might I jump from, and where would I land?

geoffharcourt:
So thoughtbot/suspenders#595 got merged. Is this still relevant, or are we accomplishing this in suspenders?

cc: @teoljungberg @gylaz

teoljungberg:
@geoffharcourt There are settings in the `ctags` file here that I think should come with the default dotfiles, but then we can customize them per project. Two things that I think should be in the default `ctags` file is:

```
--regex-ruby=/(^|[:;])[ \t]*([A-Z][[:alnum:]_]+) *=/\2/c,class,constant/
--regex-ruby=/^[ \t]*attr_(reader|writer|accessor) (:[a-z0-9_]+, )*:([a-z0-9_]+)/\3/A,attr/
```

The settings for indexing the RSpec DSL is something I can see us having on a per-repo basis

ctags
teoljungberg: Sorry for being late to respond @gabebw. 

1. Yes, that's exactly right.
2. You can navigate to them with `:tag #create` to jump to the `#create` describe-block. And also `:tag with invalid payload` to jump to those matching contexts. It's mostly useful for me to do that with either [pick] or [ctrlp]

[pick]: https://github.com/thoughtbot/pick.vim#picktag
[ctrlp]: https://github.com/ctrlpvim/ctrlp.vim

gylaz:
These are all ruby-specific. Would suspenders be a better place for these?

derekprior:
I don't know if this change is appropriate here or not, but I work on a LOT of ruby projects that don't involve suspenders.

geoffharcourt:
@teoljungberg I am definitely in favor of adding the `attr_` to our ctags. 

I'm less sold on the RSpec-specific items.

teoljungberg:
I'm good with that. Resolved in 1cc7cfd

geoffharcourt:
@teoljungberg this looks great. Awesome addition.

geoffharcourt:
Rebased and merged as d98d06a. Thanks @teoljungberg!

PR #393, opened 2015-06-10
Less annoying 81 column coloring.

lompy:
Color only those lines that are longer than 80 columns. Inspired by Damian Conway's vim setup: https://github.com/thoughtstream/Damian-Conway-s-Vim-Setup/blob/master/.vimrc#L1324

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -81,7 +81,7 @@ endif
 
 " Make it obvious where 80 characters is
 set textwidth=80
-set colorcolumn=+1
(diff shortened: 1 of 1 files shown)


gylaz:
I think it depends on the user's color scheme. I don't find my 81st column coloring bothering.

derekprior:
I prefer the persistent line so I know when it's coming.

geoffharcourt:
My color scheme is gorgeous. This isn't an issue.

PR #394, opened 2015-06-12
Add rubocop style checker to syntastic

danielnolan:
* This allows you to get the same style violation
      warnings that you would get from hound right in the left gutter of vim.
      Hound uses rubocop to check for ruby style violations. Vim is already
      configured to use syntastic. Syntastic has support for rubocop as a
      style checker. If the rubocop gem is installed the sytastic will
      enable the checker, if the rubocop gem is not found syntastic will
      just ignore the checker setting. 
* Add rubocop config from hound docs page to dotfiles.
* Configure syntastic to use rubocop as a checker in vimrc.

### DIFF
--- a/rubocop.yml
+++ b/rubocop.yml
@@ -1,0 +1,1063 @@
+AllCops:
+  Include:
+  - "**/*.gemspec"
+  - "**/*.podspec"
+  - "**/*.jbuilder"
+  - "**/*.rake"
+  - "**/*.opal"
+  - "**/Gemfile"
+  - "**/Rakefile"
+  - "**/Capfile"
+  - "**/Guardfile"
+  - "**/Podfile"
+  - "**/Thorfile"
+  - "**/Vagrantfile"
+  - "**/Berksfile"
+  - "**/Cheffile"
+  - "**/Vagabondfile"
+  Exclude:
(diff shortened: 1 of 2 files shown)


geoffharcourt:
Hey @danielnolan, I used to use Rubocop in syntastic, and I found that in large files, even on a MacBook Pro with an i7 processor, that this practice bogged down Vim. 

I think some users will decide to do Rubocop as part of their build, and some will want it in Syntastic. I think that plugins/checkers that have the potential to slow down a normally speedy Vim setup should probably be done on a per-user basis rather than in shared settings.

geoffharcourt:
For local style stuff, I have a `rubocop` rake task that runs as part of an overall `release` task that does specs, style, and security checks.

danielnolan:
Good call. I don't have too many large files, as we try to keep all classes under 100 lines. But yeah I agree this is probably more of a personal preference.

geoffharcourt:
I really see the appeal of Rubocop in-editor, but I think things that are super-easy to enable and sort of hard to disable (I think you'd have to re-specify all your Ruby Syntastic checkers to remove Rubocop, but I might be wrong) tend to be better candidates for personal customization.

PR #395, opened 2015-06-15
Ignore "useless use of variable in void context"

derekprior:
The following ERB generates a "Possibly useless use of variable in void
context" warning from Syntastic:

```erb
<%= comment %>
```

The warning can be eliminated by making it:

```erb
<%= comment.to_s %>
```

However, this behavior is the default and is redundant. There are other
cases where this warning pops up and in each case I've found it to be of
no use.


JoelQ:
Seems reasonable to me

gylaz:
:+1:

PR #397, opened 2015-06-27
Remove note about ctrl-a in tmux

CodingFu:
Since it doesn't work like that anymore.

I wonder was it changed intentionally (it didn't use to be so a year ago) or by accident?


blainesch:
Related commit

87e72a90989a5472c19daffa6055f2feb6acb909

PR #398, opened 2015-06-29
Add additional confirmation for `rm -rf`.

blainesch:


jferris:
This is an interesting idea, but I'd prefer that we don't merge it. This is a rare mistake, and every layer we add on top of base tools increases the likelihood that we introduce a bug somewhere. You also won't be able to delete the entire drive in most circumstances, and some shells (like zsh) already provide protection against mistakes like this.

PR #400, opened 2015-07-12
Update amend to set the current date on the commit.

blainesch:
When amending a commit this will update the date of the commit to be the current date.

### DIFF
--- a/bin/git-ca
+++ b/bin/git-ca
@@ -1,0 +1,1 @@
+git commit --amend -v --date=$(date +%Y-%m-%dT%H:%M:%S)
--- a/gitconfig
+++ b/gitconfig
@@ -8,7 +8,6 @@
   aa = add --all
   ap = add --patch
   branches = for-each-ref --sort=-committerdate --format=\"%(color:blue)%(authordate:relative)\t%(color:red)%(authorname)\t%(color:white)%(color:bold)%(refname:short)\" refs/remotes
-  ca = commit --amend -v
   ci = commit -v
   co = checkout
   st = status


croaky:
Cool idea. That seems like it creates it a more accurate commit. I like it.

For `sh` lines, we've been moving them out of aliases into `bin`: https://github.com/thoughtbot/dotfiles/tree/master/bin

blainesch:
@croaky I'll update the commit

blainesch:
Making a pr is difficult when are you using the dotfiles. Heh, if you reset the branch to be on `thoughtbot/master` you lose your local files.

I ended up rebasing my current branch to only have a single commit, then soft resetting it and only pushing the things necessary for this.

Updated! [:

bin/git-ca
```diff
@@ -0,0 +1 @@
+git commit --amend -v --date=$(date +%Y-%m-%dT%H:%M:%S)
```
gylaz: Thoughts on using `--date="$(date -R)"`? [Source](http://stackoverflow.com/a/9110424/373522)

blainesch: It doesn't seem to work for me.
~~~
▶ date -R
date: illegal option -- R
~~~

blainesch: Somebody also commented that `-R` doesn't work in OSX. Saying an alternative is:
~~~
git commit --amend --date="$(date +%s)"
~~~

gylaz: Apparently, it's a Linux thing. Never mind :smile:

gylaz:
Thanks @blainesch.

PR #401, opened 2015-07-18
Make Git message template more actionable

croaky:
Why this is necessary:

* All-commented template is easy to ignore/blends into Git default
  "Please enter the commit message for your changes..."
* Deleting the commented questions results in a pull request description
  that could blend the "why", "how", and "side effects" for reviewers.

How it addresses the need:

* Expose "why", "how", and "side effects" to pull request reviewers
  in a more consistent, separated, easy-to-read manner.
* Eliminates whitespace at top of commit message edit view so the person
  is more likely to "fill in the blanks" of the template structure
  rather than free-form their description at the top.

Side effects:

* None.

### DIFF
--- a/gitmessage
+++ b/gitmessage
@@ -1,20 +1,15 @@
 # 50-character subject line
 
-# 72-character wrapped description.
+Why:
 
-Why this is necessary:
-
-*
 *
 
 How it addresses the need:
 
 *
-*
 
 Side effects:
 
-*
 *
 
 # Link to the ticket, if any.
--- a/gitmessage
+++ b/gitmessage
@@ -1,4 +1,4 @@
-# 50-character subject line
+
 
 Why:
 
--- a/gitmessage
+++ b/gitmessage
@@ -1,11 +1,20 @@
-
-
 # 50-character subject line
-#
-# 72-character wrapped longer description. This should answer:
-#
-# * Why was this change necessary?
-# * How does it address the problem?
-# * Are there any side effects?
-#
-# Include a link to the ticket, if any.
+
+# 72-character wrapped description.
+
+Why this is necessary:
+
+*
+*
+
+How it addresses the need:
+
+*
+*
+
+Side effects:
+
+*
+*
+
+# Link to the ticket, if any.
--- a/gitmessage
+++ b/gitmessage
@@ -4,12 +4,4 @@ Why:
 
 *
 
-How it addresses the need:
-
-*
-
-Side effects:
-
-*
-
 # Link to the ticket, if any.


gabebw:
I am in favor.

gitmessage
```diff
@@ -1,11 +1,20 @@
...
+
+How it addresses the need:
+
+*
+*
+
+Side effects:
+
+*
+*
```
gylaz: Should these be comments? Otherwise, there is not a need to delete lines from each commit message that doesn't require some/all these lines.

JoelQ:
I like this change

gitmessage
```diff
@@ -1,11 +1,20 @@
+# 50-character subject line
 
+# 72-character wrapped description.
```
croaky: I'm not sure we need this line. The 72-character length is enforced by other mechanisms. Having the line here could be confusing?

gitmessage
```diff
@@ -1,11 +1,20 @@
...
-# 50-character subject line
-#
-# 72-character wrapped longer description. This should answer:
-#
-# * Why was this change necessary?
-# * How does it address the problem?
-# * Are there any side effects?
-#
-# Include a link to the ticket, if any.
+Why this is necessary:
```
jferris: Do you intend for users to actually include this text in the commit message, or do you expect people to rewrite or remove this?

croaky: Include. See this PR as an example.

gitmessage
croaky: Deleted the second bullet in each example so there's less to delete when editing the commit message.

croaky:
@gylaz Good to merge?

derekprior:
I'd override this in my own dotfiles. I commit dozens of times per day - usually very quickly. Having to either live with pre-populated but unused text in my interim commits or remove the text before committing would be pretty annoying.

mcmire:
I think it's a good change, but I think "How it addresses the need" is another way of saying "Why". I think in the act of explaining the purpose of the commit you'd naturally explain if there were any problems you are addressing or necessary features you're adding or whatever. But maybe that's just being nitpicky.

Also, I just noticed that the default comments that you get when you make a commit are basically saying the same thing:

```
# 50-character subject line
#
# 72-character wrapped longer description. This should answer:
#
# * Why was this change necessary?
# * How does it address the problem?
# * Are there any side effects?
```

Honestly, I think the way that these bullet points are worded make a little more sense than the headers.

croaky:
Those default comments are what's being edited in this PR. They come from .gitmessage. Agreed the "72-character" line is redundant. It's removed in this PR.

croaky:
> Having to either live with pre-populated but unused text in my interim commits or remove the text before committing would be pretty annoying.

Want to try it for a few days before deciding? I have this branch checked out and make a similar number of commits each day. I'm feeling similar friction so far, not sure I love it. Feels similar to when I started using git commit -v. Absolutely slows me down, but maybe in a good way, forces a few seconds of thought to describe what's happening, influences my opinion on whether I split the commit up well, etc.

croaky:
> I think it's a good change, but I think "How it addresses the need" is another way of saying "Why".

I've been feeling like there's a difference. Here's a few recent examples, interested to see if you agree:

* https://github.com/thoughtbot/thoughtbot.com/pull/630
* https://github.com/thoughtbot/parity/pull/45
* https://github.com/thoughtbot/parity/pull/44
* https://github.com/thoughtbot/playbook/pull/192
* https://github.com/thoughtbot/mkfriend/pull/16
* https://github.com/thoughtbot/hub/pull/181

Went off the script in https://github.com/thoughtbot/hub/pull/185

mcmire:
Okay, so you're using "Why" like "Here's some context on the issue" or "Here some context on this feature" and then "How it addresses the need" as "Here's what this commit is doing to fix the problem / implement the feature".

Maybe it seems like I'm nitpicking, but I feel like the difference between the two is not communicated well using the simple word "Why". I keep wanting to ask "why *what*", and it doesn't inspire me to explain the background behind the commit, which I think is really what it's asking. What about "Context"? I think this would also work well for your last commit, where in the first part of the message, you're essentially explaining the feature, and then you go on to list how you're doing that.

mcmire:
Also, what would be a good example of the "side effects" section? Do we anticipate this being necessary all the time?

blainesch:
@croaky only the parity repo is public.

gylaz:
I find myself getting confused on how `Why?` and `How it addresses the need` differ every time I attempt to summarize my commit. Perhaps, there is a better heading than `Why?`, or maybe combine the `Why and How`?

I also remove `Side effects:` as it's rare that I find it useful. How *deep* are we supposed to get on this one? Does my code have any side effects (e.g. mutating state)?

Overall, I like a general guideline headings like these, but I think we should iterate a bit more, and include headings for things that each commit *must* explain.

gylaz:
How about something like this, for a template:
```
# 50-character subject line

# Repeat for every problem this commit solves
* State the problem
  * State how it's addressed in the code

# Link to the ticket, if any.
```
[Example](https://github.com/thoughtbot/resuelve-tu-contabilidad/pull/283)

jferris:
> I also remove Side effects: as it's rare that I find it useful. How deep are we supposed to get on this one? Does my code have any side effects (e.g. mutating state)?

I assume this means changes behavior besides what's described in the short message. For example, if you introduce the first mailer for the application, a side effect might be that all environments now require a working SMTP server.

croaky:
Most of the time, I delete the side effects section, but when I use it, it's valuable. Yesterday it caught me from forgetting to add a redirect to Maitre-d while working on a change in robots.thoughtbot.com (different repo).

croaky:
After trying this for about two weeks, I've had the desire to simplify it a bit to:

```
# 50-character subject line

Why:

*

# Link to the ticket, if any.
```

mcmire:
Hmm... what if it were more of a prompt?

```
# -- Provide a 50-character subject line here. --

# What's the purpose of this commit?

# Is there a ticket that goes with this commit? Link to it here.
```

blainesch:
It's slowly starting to look like the original again.

MattMSumner:
Should this be left as is but make it easier to override? Currently there isn't a `.gitmessage.local` convention.

geoffharcourt:
@MattMSumner, if you create a `.gitmessage` in your own dotfiles, it will override the one in the base thoughtbot files. There's no need to do a `.gitmessage.local` because these two files aren't additive.

MattMSumner:
Makes sense. Thanks @geoffharcourt!

gylaz:
I like the fact of having either comments, or straight up text that I need to fill in, or override. I also think that a lot of folks miss the `Why` aspect of the commit message, and I'd love to incorporate that somehow into the default template.

mcmire:
> I also think that a lot of folks miss the `Why` aspect of the commit message

Okay so here's a question: why? Is it just not visible enough as a comment, does it need to *not* be a comment?

gylaz:
@mcmire I think both. A comment is just too easy to ignore, and comment at the bottom is super easy to ignore. Git adds own comments right after the title/description area, and in my case, my eyes/brain automatically ignores stuff down there.

derekprior:
Has anyone who has tried this template feel like they've changed the quality of their commits because of it. That's the heart of the question.

To me, it feels like it's a tool trying to fix something that is a people problem. But if it's working then GREAT, I guess.

croaky:
Closing without merging.

The change has been nice for me, but will probably keep evolving it in my own dotfiles. Seems better to be less invasive / prescriptive in thoughtbot's dotfiles.

mcmire:
As a followup to this PR, I've just started trying out a template that looks like this:

```


Why:

...

To satisfy the above:

...
```

I now like the "Why" header as that feels right to me. I used "To satisfy the above" because I felt like the items in the bulleted list that goes before that should be stated in imperative tense, just like the subject.

PR #404, opened 2015-08-01
Provide an example of specifying dotfiles dir

gylaz:
Why:

* There can be confusion on how to tell `rcup` where local dotfiles
  have been installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,9 @@ Install the dotfiles:
 
     env RCRC=$HOME/dotfiles/rcrc rcup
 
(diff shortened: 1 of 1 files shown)


mike-burns:
The first time through one must specify the location of the rcrc(5) file, because it is not linked yet. That's  why it's like that in the installation instructions. It might be worth mentioning that all subsequent calls can leave out the `RCRC` env var, instead.

geoffharcourt:
Made a minor edit and merged as 31a8dd3.

PR #405, opened 2015-08-05
Add ^F and ^B readline bindings and modify vi-mode binding.

jjlangholtz:
Most of the bindings for readline are already included in the zsh
configuration except for forward and backward character. For those used
to the full suite, it would be nice include ``^F`` and ``^B`` as well.

The current vi-mode mode binding of jj is troublesome for those who
either have jj in their github username or email address. One of the
most biggest problems is when copying a git repo into the terminal and
having it clipped. For example copying my repo name to git clone yields:

    git@github.com:ngholtz/dotfiles.git

In addition the buffer after typing a j means that I need to wait a
second after typing j or else it will enter vi-mode which can be a bit
annoying.

I propose that the vi-mode key just be ``^J`` to prevent both of those
issues.

Lastly, just a minor change to the rbenv init command to fix issues
where $SHELL is bin/bash causing the following message:

    /completions/rbenv.bash:14: command not found: complete

Fix suggested [here](https://github.com/sstephenson/rbenv/issues/239).

### DIFF
--- a/zshenv
+++ b/zshenv
@@ -7,7 +7,7 @@ export PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
 # load rbenv if available
 if which rbenv &>/dev/null ; then
-  eval "$(rbenv init - --no-rehash)"
+  eval "$(rbenv init - zsh --no-rehash)"
 fi
 
 # mkdir .git/safe in the root of repositories you trust
--- a/zshrc
+++ b/zshrc
@@ -45,12 +45,13 @@ unsetopt nomatch
 
 # vi mode
 bindkey -v
-bindkey "^F" vi-cmd-mode
-bindkey jj vi-cmd-mode
+bindkey "^J" vi-cmd-mode
 
 # handy keybindings
 bindkey "^A" beginning-of-line
+bindkey "^B" backward-char
 bindkey "^E" end-of-line
+bindkey "^F" forward-char
 bindkey "^K" kill-line
 bindkey "^R" history-incremental-search-backward
 bindkey "^P" history-search-backward


geoffharcourt:
One minor issue to discuss here is the use of `^B` as the new prefix key for tmux. I guess folks can press it twice, but it will be blocked for a single press for tmux users.

jjlangholtz:
Ah good point. I didn't notice that ``C-b`` is the new tmux prefix. Personally I have been using ``C-space`` for tmux to keep it similar to my vim leader ``" "``.

geoffharcourt:
@jjlangholtz, sorry to rain on your parade again, but one other conflict I see here when I was testing your proposal out is that `^j` is a common mapping used with vim-tmux-navigator so that moving down a pane or a tmux split is the same keystroke.

zshrc
```diff
@@ -45,12 +45,13 @@ unsetopt nomatch
 
 # vi mode
 bindkey -v
-bindkey "^F" vi-cmd-mode
-bindkey jj vi-cmd-mode
```
mike-burns: As the person who added this: noooooooooo!

(But really a thoughtbotter who uses these dotfiles and uses the `jj` binding should weigh in.)

geoffharcourt: As it stands, I think using `^J` here is a deal-breaker, as it clashes with vim-tmux-navigator and our vim pane navigation shortcuts. If there's another binding that doesn't conflict with core functionality I think it's worth discussing.

geoffharcourt:
Any ideas here? If we don't have a new solution that doesn't break `vim-tmux-navigator` and company I'm going to close the ticket.

croaky:
Is vim-tmux-navigator part of thoughtbot/dotfiles? I'm not up to date with that plugin and what it offers.

geoffharcourt:
Sorry, [vim-tmux-navigator[(https://github.com/christoomey/vim-tmux-navigator) isn't in our dotfiles, but our vim pane jumping shortcuts use the same mappings (and I suspect many of our dotfiles users are using it, as it's a @christoomey joint), allowing tmux users to treat vim and tmux panes in the same fashion when navigating between them. 

That means that `^h`, `^j`, `^k`, `^l` mappings are likely to already be in use in tmux, so can't be mapped for zsh. (I can press `^l` and move one pane left whether that's vim or tmux.

zshenv
```diff
@@ -7,7 +7,7 @@ export PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
 # load rbenv if available
 if which rbenv &>/dev/null ; then
-  eval "$(rbenv init - --no-rehash)"
+  eval "$(rbenv init - zsh --no-rehash)"
```
gylaz: This seems like it should be in it's own commit/PR.

gylaz:
I'm :+1: on adding the two keybindings to support readline

zshrc
gylaz: `Esc` still works. We could probably remove `jj`, and leave it up to local overrides (some people use `jk`-smash).

geoffharcourt:
@jjlangholtz, can you trim this PR down to only:

1) remove the `jj` zsh mapping 
and
2) add the '^B' and '^F' mappings

We'll do the rbenv change in a separate PR after #426 is resolved. If you don't get to it this week I'll submit a streamlined PR.

geoffharcourt:
Resolved only the `jj` binding in #432 as 844145553858fb7c8853da12a95dc0109f69bee0. Thanks for contributing!

mike-burns:
RIP `jj`.

PR #406, opened 2015-08-09
add Cocoapods alias

zinthemoney:
Similar to `bundle install`, I hope this addition to the dotfiles can save some typing.


gylaz:
This looks to be a nice candidate for `.aliases.local`.

PR #410, opened 2015-08-27
Allow to close buffer keeping the undo history

MarceloCajueiro:
Hey guys,

I don't know if it's useful for you but it's being awesome to have this undo history.

### DIFF
--- a/gitignore
+++ b/gitignore
@@ -9,3 +9,4 @@ tags
 !tags/
 tmp/**/*
 !tmp/cache/.keep
+*.un~
--- a/vimrc
+++ b/vimrc
@@ -11,6 +11,7 @@ set showcmd       " display incomplete commands
 set incsearch     " do incremental searching
 set laststatus=2  " Always display the status line
 set autowrite     " Automatically :write before running commands
+set undofile      " Allow to close buffer keeping the undo history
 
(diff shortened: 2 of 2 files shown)


BlakeWilliams:
I'm not a fan of letting programs create files in my project directories so I have the following in my `.vimrc` so I can avoid needing it in `.gitignore`:

```
if exists("+undofile")
  if isdirectory($HOME . '/.vim/undo') == 0
    :silent !mkdir -p ~/.vim/undo > /dev/null 2>&1
  endif

  set undofile
  set undodir=~/.vim/undo/
endif
```

MarceloCajueiro:
Nice. I agree with you @BlakeWilliams.

If you guys say that this patch is useful I do that improvement.

gylaz:
This sounds like a good local extension to dotfiles, but I'm not sure about this being necessary for everyone.

Let's leave this open for a whole and get some more opinions.

geoffharcourt:
I'm a fan of persistent undo history, but I had the understanding that there was a performance trade-off that not all users might want to suffer.

geoffharcourt:
I think there's enough difference of opinion on how folks want to implement undo (in the project, in their home folder, not at all) that I'm going to close the PR. 

Thank you for the idea, it looks like some folks already adopted it in their personal configuration.

PR #411, opened 2015-09-02
Change Plugin to Plug

rocLv:
Should use `Plug` other than `Plugin` to install vim plugs

### DIFF
--- a/vimrc.bundles
+++ b/vimrc.bundles
@@ -21,7 +21,7 @@ function! VundleToPlug(vundle_command, arg, ...)
   Plug a:arg, vim_plug_options
 endfunction
 
-com! -nargs=+  -bar Plugin call VundleToPlug("Plugin", <args>)
+com! -nargs=+  -bar Plugin call VundleToPlug("Plug", <args>)
 com! -nargs=+  -bar Bundle call VundleToPlug("Bundle", <args>)
 
 call plug#begin('~/.vim/bundle')


geoffharcourt:
@rocLv, this code exists as a shim for users who have pre-existing `Plugin` references from Vundle. The use of `Plugin` is intentional.

The argument `Plugin` is needed so that the user knows that `VundleToPlug()` can inform the user about their outdated reference.

rocLv:
I guess so...
It's just lots of warnings printed, and the worst is we don't know if plugins are installed.
 Maybe we should promote this.
What do you think?

blainesch:
You shouldn't be getting errors when you open vim. What is the output of rcup and the vim command giving you the errors?

rocLv:
I tried again on my Macbook Pro. There is no warning information.
But there were lots of information when I installed to my PC, which uses Ubuntu 14.04 LTS, and printed lots of messages like
```
You are using Vundle's `".a:vundle_command."` command to declare plugins. Dotfiles now uses vim-plug for plugin mangagement. Please rename uses of `".a:vundle_command."` to `Plug`. Plugin was '".a:arg."'."
```
Of cause, `a:vundle command` replaced by plugin's name.
Then I changed 
````
com! -nargs=+  -bar Plugin call VundleToPlug("Plugin", <args>)
```
to
```
com! -nargs=+  -bar Plugin call VundleToPlug("Plug", <args>)
```
and fixed the problems.

So maybe this is Ubuntu specific problem.

geoffharcourt:
@rocLv, do you know what version of vim you are using? Did you run `rcup` after changing branches (or pulling down) but before trying to open Vim?

rocLv:
@geoffharcourt Tomorrow I'll check vim version.
With `changing branches`, do you mean to change to which branch?
I installed `rcm` from git, then directly run `rcup`, and got lots of warnings.
And the plugins seemed to not be installed (got stuck) until I change `Plugin` to `Plug`.

PR #413, opened 2015-09-10
Improvements

captureteam:


gitconfig
```diff
@@ -11,16 +11,19 @@
...
-  ff = only
+  ff = false
 [commit]
   template = ~/.gitmessage
 [fetch]
   prune = true
 [rebase]
   autosquash = true
+[branch]
+  autosetuprebase = always
```
blainesch: Is there never a case when rebase wouldn't be used? I imagine in branches that you don't own, this could be harmful.

vimrc
```diff
@@ -132,12 +133,6 @@ let g:html_indent_tags = 'li\|p'
 set splitbelow
 set splitright
 
-" Quicker window movement
-nnoremap <C-j> <C-w>j
-nnoremap <C-k> <C-w>k
-nnoremap <C-h> <C-w>h
-nnoremap <C-l> <C-w>l
```
blainesch: this is useful for me, is there a reason it was removed?

vimrc.bundles
```diff
@@ -41,8 +42,10 @@ Plug 'tpope/vim-fugitive'
 Plug 'tpope/vim-rails'
 Plug 'tpope/vim-repeat'
 Plug 'tpope/vim-surround'
+Plug 'tpope/unimpaired'
 Plug 'vim-ruby/vim-ruby'
 Plug 'vim-scripts/tComment'
+Plug 'scrooloose/nerdtree'
```
blainesch: Do we need both nerdtree and ctrlp?

blainesch:
I just realized this was closed. After I made comments. Sorry, I goofed!

spidey

vimrc
bzf: It was removed to make way for the https://github.com/christoomey/vim-tmux-navigator plugin instead. :smile:

PR #414, opened 2015-09-13
Remove mention of Exuberant Ctags

andyw8:
The ctags.vim plugin was removed in #375.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ What's in it?
   HTML.
 * Use [Ag](https://github.com/ggreer/the_silver_searcher) instead of Grep when
   available.
-* Use [Exuberant Ctags](http://ctags.sourceforge.net/) for tab completion.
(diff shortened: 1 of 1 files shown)


geoffharcourt:
So we're not using the ctags.vim plugin, but Exuberant Ctags are still generating the source for autocompletion. I'm divided on whether to remove this reference.

croaky:
How about "Map <leader>ct to re-index Exuberant Ctags for the project's autocompletion." ie document the Vim settings dotfiles adds so the README acts a bit like API docs?

geoffharcourt:
Yeah, that works. @andyw8, since you waited so long to get merged, if you want to make the change I'll go ahead and merge, otherwise I'll take care of it myself on Friday.

andyw8:
@geoffharcourt updated as per @croaky's suggestion.

andyw8:
Actually, I tweaked a little, since Ctags can be used for more than autocomplete.

geoffharcourt:
Thanks for improving the documentation!

PR #415, opened 2015-09-14
Add function for opening project trello boards

graysonwright:
Also include a script for opening any tbot.io link.

These functions set up a convention for setting up tbot.io links
for project trello boards under `tbot.io/t/project-name`.

The function reads from the current directory name, and tries to open up a
tbot.io link based on that name.

I've been doing this in my dotfiles for a little while,
and it's been great.

If a tbot.io link isn't set up the first time I use it, I link it up and
it works from then on. If everyone starts setting up these links for
their projects, everyone can start benefitting from instant trello
access.

We currently have set up:

- t/administrate
- t/apprentice
- t/blog
- t/feed
- t/formkeep
- t/growth
- t/hound
- t/success
- t/thoughtbot
- t/upcase
- t/upcase-content
- t/upcase-growth
- t/weekly-iteration

### DIFF
--- a/zsh/functions/tbot
+++ b/zsh/functions/tbot
@@ -1,0 +1,3 @@
+tbot () {
+  open "https://tbot.io/$1"
+}
--- a/zsh/functions/trello
+++ b/zsh/functions/trello
@@ -1,0 +1,3 @@
+trello () {
+  tbot "t/$(basename $PWD)"
+}


zsh/functions/tbot
```diff
@@ -0,0 +1,3 @@
+tbot () {
+  open https://tbot.io/$1
```
gabebw: The URL should be quoted.

teoljungberg: What happends if the directory has a space in it?

zsh/functions/trello
```diff
@@ -0,0 +1,3 @@
+trello () {
+  tbot t/$(basename $PWD)
```
gabebw: The argument to tbot should be quoted.

jferris:
The standard for opening a file or URL on Linux is [`xdg-open`](http://linux.die.net/man/1/xdg-open).

gabebw:
This is really specific to thoughtbot.

Should everything in the dotfiles be of general use, or is this more about building something for us that other people can take or leave?

mike-burns:
@gabebw  This is the thoughtbot dotfiles.

mike-burns:
Why is this a function?

gylaz:
This seems like this needs some trying out amongst our teammates before we decide to merge this in. I'd like to not merge arbitrary utilities into our dotfiles.

mike-burns:
On Freedesktop setups it's xdg-open(1), too. I could see something like:

```sh
browser() {
  if [ "x$BROWSER" = "x" ]; then
  if command xdg-open; then
    echo xdg-open
  elif command open; then
    echo open
  else
    echo echo
  fi
  else
    echo "$BROWSER"
  fi
}

tbot {
  `browser` "http://tbot.io/$1"
}
```

Or something. Hopefully something better.

graysonwright:
Alright, I created a research card for this. https://trello.com/c/0FrpZhgd/555-command-line-tool-for-opening-trello-boards

We can get a few people using this for a couple weeks, then see if we'd like to merge it in. If you're interested, put your face on the card!

geoffharcourt:
I'd be more in favor of this if it were some kind of private Homebrew forumula (even if it's not really private, just something unadvertised externally) rather than in our dotfiles. It looks extremely useful, but thus far our dotfiles have always been things anyone could use (I certainly made use of them before I came aboard).

graysonwright:
That reasoning makes sense.

geoffharcourt:
@graysonwright I am extremely interested in the results of this research if it results in a Homebrew formula. If you need a pair I'd love to work on it with you.

PR #416, opened 2015-09-24
Allow local binstub overrides

derekprior:
We're now checking in binstubs to each project's `bin` directory and,
optionally, adding that `bin` directory to our path by adding the
`.git/safe` directory.

There are some environments where we need to tweak the behavior of the
binstub locally. For instance, when running docker we don't want to
execute the binstub locally; rather, we want to execute that binstub in
the context of our docker container.

On a current project that uses Docker, I have the following `rails`
binstub in my `local_bin` directory:

```
#!/bin/sh

docker-compose run web rails "$@"
```

By putting this command first in my path, it allows all of my other
rails aliases to work as they would if I were working locally. I have
similar commands for `rspec` and `rake`.

### DIFF
--- a/gitignore
+++ b/gitignore
@@ -9,3 +9,4 @@ tags
 !tags/
 tmp/**/*
 !tmp/cache/.keep
+local_bin/
--- a/zshenv
+++ b/zshenv
@@ -11,7 +11,7 @@ if which rbenv &>/dev/null ; then
 fi
 
 # mkdir .git/safe in the root of repositories you trust
-export PATH=".git/safe/../../bin:$PATH"
+export PATH=".git/safe/../../local_bin:.git/safe/../../bin:$PATH"
 
 # Local config
 [[ -f ~/.zshenv.local ]] && source ~/.zshenv.local


gitignore
```diff
@@ -9,3 +9,4 @@ tags
 !tags/
 tmp/**/*
 !tmp/cache/.keep
+local_bin/
```
gylaz: Does it not make sense to commit `local_bin/`?

derekprior: I could go either way. I checked it in on the docker project I was on, but it made me uneasy - maybe because the name is "*local*_bin".

jferris: What do you think about committing these as `host_bin`?

geoffharcourt:
@derekprior, I'm waiting for us to figure out our resolution on #426 before we move forward here.

geoffharcourt:
@derekprior now that we've sorted out some larger zsh issues, I'm eager to get this merged in. Have you come to a conclusion about how you want to name the local binstub directory per the above discussion?

croaky:
Cool idea, I like it.

derekprior:
I can't say I'm too interested in this right now. I didn't particularly enjoy working in the docker environment and I', not doing it currently. I don't care enough to continue the charge for this PR.

Anyone want to make the suggested improvements and merge it? I'm now indifferent.

geoffharcourt:
@derekprior unless someone is going to be using this in an ongoing basis it might be best to close. If we have folks using this with docker (or in similar situations) later, we can always reopen it.

Apologies for the delay after you initially suggested it, there was a lot of churn doing the OS X El Cap update and I didn't want to add further uncertainty.

jferris:
I'm still interested in this, but I'd also be fine submitting a new pull request once I've been able to play with it more thoroughly.

geoffharcourt:
OK, cool. I'll leave it open while you're taking a look.

derekprior:
I'm going to close this. It's not complex and I got the idea from Joe to begin with. He can recreate when he's ready.

PR #418, opened 2015-10-01
Added common configurations

mwenger1:
The goal is to try to standarize the configurations the whole team is on board with to reduces friction when pairing.

Feel free to flag anything that you don't think it a good fit for all developers to have on their machine.


gylaz:
This seems like a lot of changes at once. Did you mean to open this PR against upstream?

mwenger1:
Sorry about that. You're exactly right. It was an accident.

Issue #420, opened 2015-10-02
$PATH on OS X El Capitan


ventsislaf:
It looks like `path_helper` on OS X El Capitan is moved from `/etc/zshenv` to `/etc/zprofile` which is overriding the `$PATH` after the user's customizations.
More info: http://www.zsh.org/mla/users/2015/msg00727.html

I'm not sure if this has to be addressed in [thoughtbot/dotfiles](https://github.com/thoughtbot/dotfiles) or [thoughtbot/laptop](https://github.com/thoughtbot/laptop).

ventsislaf:
As a fast workaround I moved this code

```sh
# ensure dotfiles bin directory is loaded first
export PATH="$HOME/.bin:/usr/local/sbin:$PATH"

# load rbenv if available
if which rbenv &>/dev/null ; then
  eval "$(rbenv init - --no-rehash)"
fi

# mkdir .git/safe in the root of repositories you trust
export PATH=".git/safe/../../bin:$PATH"
```

from `~/.zshenv` to `~/.zprofile`.

geoffharcourt:
@ventsislaf, I'm seeing the same thing. Looking at a solution that will work across Yosemite and El Capitan.

geoffharcourt:
Fixed in 15de0bcfee0d7a6472da048029408c5faedbae4e through documentation. Thank you for bringing this to our attention.

PR #421, opened 2015-10-02
Fix $PATH on OS X El Capitan

reshleman:
OS X El Capitan [moved `path_helper`][1] from `/etc/zshenv` to
`/etc/zprofile`. This causes the changes to `$PATH` in `~/.zshenv` to be
overridden/re-ordered.

Moving the `$PATH` modifications to `~/.zprofile` causes them to be
loaded after OS X's `path_helper`, resulting in the expected order.

Per convention, `~/.zprofile` sources an optional `~/.zprofile.local`.

Additionally, because `tmux` loads `zsh` as a login shell, `~/.zprofile`
is sourced twice; this commit adds a command to `~/.zprofile` to remove
the resulting duplicate entries from `$PATH`.

[1]: http://www.zsh.org/mla/users/2015/msg00726.html

### DIFF
--- a/zprofile
+++ b/zprofile
@@ -1,0 +1,17 @@
+# ensure dotfiles bin directory is loaded first
+export PATH="$HOME/.bin:/usr/local/sbin:$PATH"
+
+# load rbenv if available
+if which rbenv &>/dev/null ; then
+  eval "$(rbenv init - --no-rehash)"
+fi
+
+# mkdir .git/safe in the root of repositories you trust
+export PATH=".git/safe/../../bin:$PATH"
+
+# Local config
+[[ -f ~/.zprofile.local ]] && source ~/.zprofile.local
+
+# tmux loads zsh as a login shell, which may result in duplicate $PATH entries
+# from ~/.zprofile. This removes duplicates from $PATH.
+typeset -U PATH
--- a/zshenv
+++ b/zshenv
@@ -2,16 +2,5 @@
 export VISUAL=vim
 export EDITOR=$VISUAL
 
-# ensure dotfiles bin directory is loaded first
-export PATH="$HOME/.bin:/usr/local/sbin:$PATH"
-
-# load rbenv if available
-if which rbenv &>/dev/null ; then
-  eval "$(rbenv init - --no-rehash)"
-fi
-
-# mkdir .git/safe in the root of repositories you trust
-export PATH=".git/safe/../../bin:$PATH"
-
 # Local config
 [[ -f ~/.zshenv.local ]] && source ~/.zshenv.local


reshleman:
Whelp, wish I had seen #420 before I spent time troubleshooting this. Thanks @ventsislaf!

Would probably be good to get a few more sets of eyes on this.

geoffharcourt:
@reshleman, I ran `rcup` after checking out this branch on my El Capitan machine and it worked. I have a minor concern that if any users have taken the [advice from the zsh users group](http://www.zsh.org/mla/users/2015/msg00727.html) and deleted `/etc/zprofile` or commented out the `path_helper` call that this code won't work.

geoffharcourt:
I'm leaning toward explicitly defining `PATH` in `.zprofile`, so that the code works whether or not the user has intervened with `path_helper`.

geoffharcourt:
@reshleman, I think there's some remaining path reordering to be done. With base dotfiles and no the default El Capitan `path_finder` unmodified, I end up with this path:

```shell
% echo $PATH
/Users/geoff/.rbenv/shims:/Users/geoff/.bin:.git/safe/../../bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
```

Do you have something different? The `.git/safe/../../bin` binstubs folder should be before the rbenv shims, or else binstubs won't get called.

geoffharcourt:
A second problem I see here is that by moving this code to `.zprofile`, it means that scripts and non-login shells won't have access to rbenv, and may run the wrong version of Ruby. I think we have to keep the path configuration in `.zshenv`, explicitly set paths there, and add instructions for users to comment out the `path_helper` call in `/etc/zprofile`.

Shades of the long discussion here: https://github.com/thoughtbot/dotfiles/pull/307, but the difference is that by moving the path mangling later, Apple is making it harder to just override their changes.

cc: @thoughtbot/shell

reshleman:
@geoffharcourt With the unmodified El Capitan `path_finder` I get:

On `master`:

```
% path
/usr/local/bin
/usr/bin
/bin
/usr/sbin
/sbin
.git/safe/../../bin
/Users/reshleman/.bin
/usr/local/sbin
```

On this branch:

```
% path
.git/safe/../../bin
/Users/reshleman/.rbenv/shims
/Users/reshleman/.bin
/usr/local/sbin
/usr/local/bin
/usr/bin
/bin
/usr/sbin
/sbin
```

geoffharcourt:
OK, that looks fine, thanks for showing your path.

I think the issue of doing this loading in `.zprofile` is going to break non-login shells, however. I think we need to add instructions remove `/etc/zprofile` and keep this path loading code in `~/.zshenv` where zsh suggests it live.

geoffharcourt:
Check out this comment from the last time OS X broke zsh paths: https://github.com/thoughtbot/dotfiles/pull/307#issuecomment-57826349

reshleman:
> I think the issue of doing this loading in .zprofile is going to break non-login shells, however.

Yeah, that's a good point. I hadn't considered that.

> add instructions for users to comment out the `path_helper` call in `/etc/zprofile`.

I actually tried this originally, and it didn't resolve the problem for me. Commenting out the `path_helper` line (or removing `/etc/zprofile`), using `master` dotfiles gives me:

```
% path
.git/safe/../../bin
/Users/reshleman/.bin
/usr/local/sbin
/usr/bin
/bin
```

It looks like removing `path_helper` prevents `/etc/paths` from being loaded, so `/usr/local/bin` is missing; that means nothing installed via Homebrew (including `rbenv`) is on `$PATH`. 

Are you getting a different result after removing the `path_helper` line?

geoffharcourt:
No. Here's how I think we should fix this:

1. Revert all the changes here.
2. Modify `zshenv` so that `/usr/local/bin` is added to the path just after `/usr/local/sbin` [here](https://github.com/thoughtbot/dotfiles/blob/master/zshenv#L6)
3. Add an instruction advising El Capitan users to rename `/etc/zprofile` to `/etc/zprofile.old`.

```
OS X El Capitan (10.11) edits the path from `/etc/zprofile`, which gets sourced  
*after* `.zshenv`, and will reorder your path in such a way that Homebrew and    
rbenv may not work. You can rename the file to prevent it from mangling your     
environment's path:                                                              
                                                                                    
  % sudo mv /etc/zprofile /etc/zprofile.old  
```
You've already done so much legwork here, do you want to author the commit with these three steps? I'm happy to do it if you've had enough wrangling with this today.

geoffharcourt:
After internal discussion, there's a cleaner solution from @derekprior that doesn't discard the benefits of `path_helper`. in #307, the discussion concluded that keeping `path_helper` was important for other reasons such as installed software adding paths for users.

I'm going to update the README to instruct users to move `/etc/zprofile` to be `/etc/zshenv`, and then El Capitan will behave like Yosemite when loading the zsh environment,

reshleman:
@geoffharcourt Got it -- thanks for working through this with me.

geoffharcourt:
Thank you for investigating! This was a much deeper hole to dig through than I had expected it to be.

PR #423, opened 2015-10-02
Warn users with harmful `path_helper` calls

derekprior:
El Capitan moves the `path_helper` call to `zprofile`, but our dotfiles
had already accounted for it being in `zshenv` where it was previously.

### DIFF
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -9,7 +9,7 @@ fi
 
 vim -u $HOME/.vimrc.bundles +PlugInstall +PlugClean! +qa
 
-if grep --quiet --no-messages 'path_helper' /etc/zprofile; then
+if grep --quiet --no-messages path_helper /etc/zprofile; then
   cat <<eom
 
 Your system is sourcing the OS X 'path_helper' from /etc/zprofile.
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -6,4 +6,15 @@ if [ ! -e $HOME/.vim/autoload/plug.vim ]; then
   curl -fLo $HOME/.vim/autoload/plug.vim --create-dirs \
       https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 fi
+
 vim -u $HOME/.vimrc.bundles +PlugInstall +PlugClean! +qa
+
+if grep --quiet --silent 'path_helper' /etc/zprofile; then
+  cat <<eom
+
+Your system is sourcing the OS X `path_helper` from /etc/zprofile.
+This will interfere with path ordering. For more information, see:
+https://github.com/thoughtbot/dotfiles#note-for-os-x-el-capitan-users
+
+eom
+fi
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -9,7 +9,7 @@ fi
 
 vim -u $HOME/.vimrc.bundles +PlugInstall +PlugClean! +qa
 
-if grep --quiet --silent 'path_helper' /etc/zprofile; then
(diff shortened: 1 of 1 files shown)


geoffharcourt:
Correctly warned on my machine.

geoffharcourt:
The warning message is missing the word `path_helper` on my screen output.

derekprior:
That's because I was so helpfully calling path_helper rather than displaying it.

geoffharcourt:
One more issue: if `zprofile` doesn't exist, there's an error message:

```shell
❯ rcup
grep: /etc/zprofile: No such file or directory
```

derekprior:
hm. I thought that's what `--silet` was supposed to fix.

hooks/post-up
```diff
@@ -6,4 +6,15 @@ if [ ! -e $HOME/.vim/autoload/plug.vim ]; then
   curl -fLo $HOME/.vim/autoload/plug.vim --create-dirs \
       https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 fi
+
 vim -u $HOME/.vimrc.bundles +PlugInstall +PlugClean! +qa
+
+if grep --quiet --silent 'path_helper' /etc/zprofile; then
```
gabebw: According to `man grep` on OS X, `--quiet` and `--silent` do the same thing:

```
 -q, --quiet, --silent
             Quiet mode: suppress normal output.  grep will only search a file until a match has been found, making searches potentially less expensive.
```

Do you mean `--quiet --no-messages`?

```
     -s, --no-messages
             Silent mode.  Nonexistent and unreadable files are ignored (i.e. their error messages are suppressed).
```

Is there a difference between OS X, *BSD, and Linux grep here? @thoughtbot/shell

derekprior:
Updated. The flag I wanted was `--no-messages`

hooks/post-up
```diff
@@ -6,4 +6,15 @@ if [ ! -e $HOME/.vim/autoload/plug.vim ]; then
   curl -fLo $HOME/.vim/autoload/plug.vim --create-dirs \
       https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 fi
+
 vim -u $HOME/.vimrc.bundles +PlugInstall +PlugClean! +qa
+
+if grep --quiet --no-messages 'path_helper' /etc/zprofile; then
```
gabebw: Technically this doesn't require the quotes. But it doesn't really matter.

gabebw:
I have no further input on the quality of this shell script.

hooks/post-up
```diff
@@ -6,4 +6,15 @@ if [ ! -e $HOME/.vim/autoload/plug.vim ]; then
   curl -fLo $HOME/.vim/autoload/plug.vim --create-dirs \
       https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 fi
+
 vim -u $HOME/.vimrc.bundles +PlugInstall +PlugClean! +qa
+
+if grep --quiet --no-messages path_helper /etc/zprofile; then
```
mike-burns: The more portable way to write this is:

```sh
if grep path_helper /etc/zprofile >/dev/null 2>/dev/null; then
```

That will work on unix systems in addition to GNU.

derekprior:
Looks like we're going in another direction.

geoffharcourt:
Hi @derekprior I wanted to wait until we resolved the issues in #426 before making a call here.

derekprior:
In any event - the code there will be more useful in displaying a better message.

PR #424, opened 2015-10-03
Fix a command in README

zillou:
The quoted instruction to move `/etc/zprofile` to `/etc/zshenv` has  space between the old and new filenames which breaks in the shell..

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,16 +45,16 @@ You can safely run `rcup` multiple times to update:
 You should run `rcup` after pulling a new version of the repository to symlink
 any new files in the repository.
 
-#### Note for OS X El Capitan Users 
+#### Note for OS X El Capitan Users
 
-OS X El Capitan (10.11) calls `path_helper` from `/etc/zprofile`, which gets 
-sourced *after* `.zshenv`, and will reorder your path in such a way that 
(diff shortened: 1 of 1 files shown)


ventsislaf:
@geoffharcourt do you think it's a good idea to automate this in [thoughtbot/laptop](https://github.com/thoughtbot/laptop)?

geoffharcourt:
I'm thinking about this, but leaning toward no further intervention.

Laptop without dotfiles gives you a good vanilla setup to start doing Ruby development without much investment in console-based editors, etc. There we are making the bare minimum changes to zsh configuration to help users get started. Laptop works out of the box (the path changes only become a problem when we introduce more zsh environment customization with dotfiles), and I don't want to introduce something that modifies a user's environment in an unexpected way if they can be productive with Laptop as it currently functions.

I wish Apple wouldn't change these little zsh environment quirks with each release.

ventsislaf:
Makes sense.

> I wish Apple wouldn't change these little zsh environment quirks with each release.

I can't agree more on this one.

geoffharcourt:
Yeah, I'm not totally convinced we shouldn't rename it or move it, but I want to be patient. A zsh users list I read advised deleting it, which turns out to deny use of paths defined in `paths.d`.

PR #425, opened 2015-10-03
Use vim-projectionist (and vim-rake)

### LINKED ISSUE #381: Alternate between spec and lib files in Vim?
When I'm working in a gem, I don't have the nice `:A` I'm used to from Rails.vim, alternating between spec/models/user_spec.rb to app/models/user.rb, for example.

In a gem, we pretty regularly have `spec` and `lib` directories. Wondering if anyone has a favorite setup for this scenario? I see https://github.com/clembradley/alt-ruby as maybe having potential.

geoffharcourt:
Use vim-projectionist and vim-rake

[vim-projectionist](https://github.com/tpope/vim-projectionist) provides vim-rails-like shortcuts for alternate and
related files. It allows users to provide a custom set of "projections"
that map filetypes to their alternate and related counterparts. Custom
mappings can be declared in a JSON-like syntax.

This change provides authors of gems and other non-Rails Ruby projects
with access to the alternate file `:A` shortcuts to navigate between
code and specs. [vim-rake](https://github.com/tpope/vim-rake) automatically maps code/spec mappings for Ruby
files if vim-projectionist is available.

Close #381.


jferris:
Looks good to me.

gylaz:
I'm :+1: but just so I understand, this is already handled by `vim-rails` for Rails projects, right?

MattMSumner:
:+1:

geoffharcourt:
@gyglaz, correct. We don't have this functionality for gems and other non-Rails projects.

PR #428, opened 2015-10-13
Use `command -v` instead of `which` in conditionals

codenamev:
Reason: https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script/677212#677212

Primarily avoid using an external process for simple checks like this, and an added side-effect of consistency across OSes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Your `~/.vimrc.local` might look like this:
 Your `~/.zshenv.local` might look like this:
 
     # load pyenv if available
-    if which pyenv &>/dev/null ; then
+    if command -v pyenv &>/dev/null ; then
       eval "$(pyenv init -)"
     fi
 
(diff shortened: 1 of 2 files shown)


jferris:
In the Stack Overflow question you linked, it looks like the author recommends `command -v` rather than `type` for POSIX.

codenamev:
@jferris updated.

jferris:
@codenamev thanks!

I tried this out locally (in zsh) and it seems to work the way we expect.

@thoughtbot/shell - any thoughts on this?

mike-burns:
Works for me.

gabebw:
`command -v` works in OS X -- it's my standard way to check if something is available.

mike-burns:
Oh, and I also tried it on ksh and zsh and can confirm that it works.

geoffharcourt:
Worked for me. Rebased and merged as 9aa2cd5. Thanks for the contribution!

PR #430, opened 2015-10-15
Split zsh configuration into smaller files

mike-burns:
This allows people to incorporate the thoughtbot dotfiles into their own
dotfiles in a more fine-grained manner.

I left some things in zshrc that we should eventually handle more
precisely:

- Load `.zsh/functions/*`. This could instead be replaced with: `mv
  .zsh/functions/* .zsh/configs`.
- Load `.aliases`. This could instead be replaced with: `mv .aliases
  .zsh/configs/aliases.zsh`.
- Load `.zshrc.local`. This file can realistically go away entirely,
  with people adding their own files to `.zsh/configs`.

A further refactoring, which I have done locally, is to introduce a
`~/.sh/configs` directory, in which people can put POSIX-specific
configuration that can be shared between GNU Bash, zsh, ksh, etc:
aliases, functions, paths, prompts, and so on. But one step at a time.

### DIFF
--- a/zsh/completion/_g
+++ b/zsh/completion/_g
@@ -1,0 +1,2 @@
+#compdef g
+compdef g=git
--- a/zsh/configs/color.zsh
+++ b/zsh/configs/color.zsh
@@ -1,0 +1,6 @@
+# makes color constants available
+autoload -U colors
+colors
+
+# enable colored output from ls, etc. on FreeBSD-based systems
+export CLICOLOR=1
--- a/zsh/configs/editor.zsh
+++ b/zsh/configs/editor.zsh
@@ -1,0 +1,2 @@
+export VISUAL=vim
+export EDITOR=$VISUAL
--- a/zsh/configs/history.zsh
+++ b/zsh/configs/history.zsh
@@ -1,0 +1,4 @@
+setopt hist_ignore_all_dups inc_append_history
+HISTFILE=~/.zhistory
+HISTSIZE=4096
+SAVEHIST=4096
--- a/zsh/configs/keybindings.zsh
+++ b/zsh/configs/keybindings.zsh
@@ -1,0 +1,13 @@
+# vi mode
+bindkey -v
+bindkey "^F" vi-cmd-mode
+
+# handy keybindings
+bindkey "^A" beginning-of-line
+bindkey "^E" end-of-line
+bindkey "^K" kill-line
+bindkey "^R" history-incremental-search-backward
+bindkey "^P" history-search-backward
+bindkey "^Y" accept-and-hold
+bindkey "^N" insert-last-word
+bindkey -s "^T" "^[Isudo ^[A" # "t" for "toughguy"
--- a/zsh/configs/options.zsh
+++ b/zsh/configs/options.zsh
@@ -1,0 +1,9 @@
+# awesome cd movements from zshkit
+setopt autocd autopushd pushdminus pushdsilent pushdtohome cdablevars
+DIRSTACKSIZE=5
+
+# Enable extended globbing
+setopt extendedglob
+
+# Allow [ or ] whereever you want
+unsetopt nomatch
--- a/zsh/configs/post/completion.zsh
+++ b/zsh/configs/post/completion.zsh
@@ -1,0 +1,6 @@
+# load our own completion functions
+fpath=(~/.zsh/completion /usr/local/share/zsh/site-functions $fpath)
+
+# completion
+autoload -U compinit
+compinit
--- a/zsh/configs/post/path.zsh
+++ b/zsh/configs/post/path.zsh
@@ -1,0 +1,10 @@
+# ensure dotfiles bin directory is loaded first
+export PATH="$HOME/.bin:/usr/local/sbin:$PATH"
+
+# load rbenv if available
+if command -v rbenv >/dev/null; then
+  eval "$(rbenv init - --no-rehash)"
+fi
+
+# mkdir .git/safe in the root of repositories you trust
+export PATH=".git/safe/../../bin:$PATH"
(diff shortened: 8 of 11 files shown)


bernerdschaefer:
This looks great. Does the README need any updates? It looks like [we recommend][zsh-config] users put their own customizations into `~/.zsh/configs/*`.

[zsh-config]: https://github.com/thoughtbot/dotfiles#zsh-configurations

mike-burns:
@bernerdschaefer the README shouldn't need any updates for this, since we already recommend `.zsh/configs`.

geoffharcourt:
@mike-burns this looks really cool. I think our zsh configuration had grown enough to warrant the change. One question: have you had issues with the order these settings are loaded? I'm assuming that files in `.zsh/configs` get loaded alphabetically.

mike-burns:
The files in `.zsh/configs` are loaded in i-node order, most likely -- whatever `readdir(3)` returns. So far as I know, the order for these config files do not matter. If they do, we can use the `.zsh/configs/pre` (or `post`) directory to manage it.

geoffharcourt:
I am very much in favor of changes that allow wider adoption without forcing every part of our configuration on users. I've noticed that there are lots of folks who don't actually use our dotfiles, but follow them and incorporate some of the changes (or even make contributions!), which seems inefficient.

I want to give folks a little bit of time to respond to this, but unless there are meaningful objections I plan to merge it as is.

geoffharcourt:
@mike-burns we had some config move into `zshrc` from `zshenv` in the last bunch of dotfiles contributions. I'm happy to move these over and to reconcile the changes unless you want to do it yourself.

geoffharcourt:
@mike-burns: I rebased and reconciled some changes that had been made in the period since you opened the PR. The most significant change is the addition of `zsh/path.zsh`, which maintains the move we made in 8fbb6f291840693342b18940b10df9cd024c715b. I had thought of potentially creating `ruby.zsh`, to handle the rbenv invocation (making swapping in RVM or avoiding the use of rbenv altogether an easy personal customization), but it's position in the order of `PATH` changes is important.

The `zsh/path.zsh` addition was made under your name, so I wanted to get your feedback before moving ahead.

geoffharcourt:
One other issue we need to resolve: in https://github.com/thoughtbot/dotfiles/blob/31a8dd34f0816b43f8fa93faab635e6dde099217/zsh/functions/g after defining our `g()` function, we declare that `g` uses `git`'s completion. `compdef` is not loaded yet, so this generates an error (and fails to enable command-completion on git commands for `g`). 

Should we be running `compdef g=git` in `zsh/configs/post`?

mike-burns:
I've fixed 89b60fcb49baf2228ab3e5c27be65312ce4b9b8d in #439.

mike-burns:
Re PATHs: I don't understand what you all are doing around PATHs and I don't know how to help. I also don't use rbenv, OS X, or `.git/safe/../../bin`.

But I think the solution is: put the `.git/safe/../../bin` bit in a `post/` subdirectory.

(Also, while there: you don't need to export PATH -- in fact, it's probably a bug if child programs use your shell's PATH variable.)

derekprior:
> But I think the solution is: put the .git/safe/../../bin bit in a post/ subdirectory.

This falls down when there's a second thing we want to have happen near the end of PATH because we can't control the ordering and maintain topical separation.

For instance, I'd want to have a `post` file for ruby version manager (and any other version manager that operates by mucking with PATH) and for the `.git/safe` convention for trusting local bin stubs. But the order of those is important.

I think we can live with having a single file for manipulating PATH in post, but it's something we don't have to worry about with the current setup.

derekprior:
Having written the above, I immediately realized that my mistake was thinking of the topical separation from the point of view of the side effects. In this case: "Load ruby version manager, use local binstubs". I think what's best about this approach here is that the topical separation is along the ZSH functionality axis. completeion,settings, etc. In this context, `paths` makes sense in `post`. 

It's also probably why this approach will work a lot better than my previous attempts at topical separation.

geoffharcourt:
So the rule here would be only one `PATH` modification in post? That seems reasonable to me.

zsh/configs/completion.zsh
```diff
@@ -0,0 +1,6 @@
+# load our own completion functions
+fpath=(~/.zsh/completion /usr/local/share/zsh/site-functions $fpath)
+
+# completion
+autoload -U compinit
+compinit
```
geoffharcourt: Any thought to adding this to `zsh/configs/post/completion.zsh`? Then any completions defined on functions or not-yet-loaded commands won't generate an error.

geoffharcourt:
I think autocompletion for commands we set is broken with the splitup of the files. I've been reading documentation on `compinit` to see exactly why. We have some overrides for some commands like `ag` and `production`. 

I tried moving `zsh/config/completion.zsh` to `zsh/configs/post` to see if we had a load order issue with no success.

geoffharcourt:
@mike-burns I made a few updates that can be squashed in. I'm going to work with this setup for a few days, but I like the organization and the potential to let users pick and choose the pieces of our setup they want to use in a more a la carte fashion.

geoffharcourt:
@mike-burns I've been working with the config files split up, and with the changes I've squashed it's been working well. Wanted to get your eyes on it before I commit the rebased changes.

zsh/completion/_g
```diff
@@ -0,0 +1,2 @@
+#compdef g
+compdef g=git
```
mike-burns: Nice.

geoffharcourt: This only works post-split, our old setup sourced functions before we loaded `compinit`.

mike-burns:
@geoffharcourt beautiful. LGTM.

geoffharcourt:
@mike-burns this was an awesome change. I think it will let many more people use the dotfiles both within and outside of thoughtbot, allowing for much more _a la carte_ choosing of what parts of the setup to incorporate into personal configurations.

Thank you very much for the idea and the implementation.

PR #431, opened 2015-10-16
Add `git pf` alias

seanpdoyle:
Using `--force-with-lease` allows one to force push without the risk of
unintentionally overwriting someone else's work.

The git-push(1) man page states:

> Usually, "git push" refuses to update a remote ref that is not an
> ancestor of the local ref used to overwrite it.
>
> This option overrides this restriction if the current value of the
> remote ref is the expected value. "git push" fails otherwise.
>
> --force-with-lease alone, without specifying the details, will protect
> all remote refs that are going to be updated by requiring their
> current value to be the same as the remote-tracking branch we have for
> them.
>
> --@calleerlandsson
> -- https://github.com/thoughtbot/guides/pull/363

### DIFF
--- a/gitconfig
+++ b/gitconfig
@@ -10,6 +10,7 @@
   branches = for-each-ref --sort=-committerdate --format=\"%(color:blue)%(authordate:relative)\t%(color:red)%(authorname)\t%(color:white)%(color:bold)%(refname:short)\" refs/remotes
   ci = commit -v
   co = checkout
+  pf = push --force-with-lease
   st = status
 [core]
   excludesfile = ~/.gitignore


mike-burns:
I'm happy with this.

geoffharcourt:
This is fantastic. Looks good in my testing.

jferris:
Nice.

geoffharcourt:
@seanpdoyle, as this doesn't impact anyone's existing git operations and just layers on top, I think this is safe to rebase and merge. I'm going to use this all the time.

seanpdoyle:
Merged in https://github.com/thoughtbot/dotfiles/commit/7733711c3b82d2ead7680c20b42d89dfcb2230bf

geoffharcourt:
:clap:

PR #433, opened 2015-10-19
Allow autoloading of proprietary binfiles

TheLonelyGhost:
Sometimes proprietary scripts specific to server administration are required, but don't fit in the same repository as the rest of the source code. This adds all subdirectories of `~/.proprietary` that contain a directory called "bin" to your path automatically, allowing specific projects' proprietary scripts to be split up into separate directories. Example is in a comment in the code.

Side effects are that the project-specific bin files are included globally. Eventually it would be nice to include based on current working directory, but that borders on [`direnv`](https://github.com/direnv/direnv)'s territory.

### DIFF
--- a/zshenv
+++ b/zshenv
@@ -27,3 +27,16 @@ MSG
 fi
 
 unset _old_path
+
+# load proprietary scripts into path
+if [ -d "$HOME/.proprietary" ]; then
+  # add all sudirectories containing a directory named 'bin' to PATH
+  # EXAMPLE: '~/.proprietary/derp/bin' and '~/.proprietary/bin' would be added,
+  #          but '~/.proprietary/foo/bar/bin' would not.
+  PATH="$(find -L $HOME/.proprietary -maxdepth 2 -mindepth 1 -name 'bin' -type d | tr '\n' ':')$PATH"
+fi
+
+# load rbenv if available
+if command -v rbenv &>/dev/null ; then
+  eval "$(rbenv init - --no-rehash)"
+fi
--- a/zshenv
+++ b/zshenv
@@ -28,20 +28,4 @@ fi
 
 unset _old_path
 
-# load proprietary scripts into path
-if [ -d "$HOME/.proprietary" ]; then
-  # add all sudirectories containing a directory named 'bin' to PATH
-  # EXAMPLE: '~/.proprietary/derp/bin' and '~/.proprietary/bin' would be added,
-  #          but '~/.proprietary/foo/bar/bin' would not.
-  PATH="$(find -L $HOME/.proprietary -maxdepth 2 -mindepth 1 -name 'bin' -type d | tr '\n' ':')$PATH"
-fi
-
-# load rbenv if available
-if command -v rbenv &>/dev/null ; then
-  eval "$(rbenv init - --no-rehash)"
-fi
-
-# mkdir .git/safe in the root of repositories you trust
-PATH=".git/safe/../../bin:$PATH"
-
 export PATH
--- a/zshenv
+++ b/zshenv
@@ -40,3 +40,8 @@ fi
 if command -v rbenv &>/dev/null ; then
   eval "$(rbenv init - --no-rehash)"
 fi
+
+# mkdir .git/safe in the root of repositories you trust
+PATH=".git/safe/../../bin:$PATH"
+
+export PATH
--- a/zshrc
+++ b/zshrc
@@ -3,7 +3,7 @@ export VISUAL=vim
 export EDITOR=$VISUAL
 
 # ensure dotfiles bin directory is loaded first
-export PATH="$HOME/.bin:/usr/local/sbin:$PATH"
+PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
 # load rbenv if available
 if command -v rbenv >/dev/null; then
@@ -11,7 +11,15 @@ if command -v rbenv >/dev/null; then
 fi
 
 # mkdir .git/safe in the root of repositories you trust
-export PATH=".git/safe/../../bin:$PATH"
+PATH=".git/safe/../../bin:$PATH"
+
+# load proprietary scripts into path
+if [ -d "$HOME/.proprietary" ]; then
+  # add all sudirectories containing a directory named 'bin' to PATH
+  # EXAMPLE: '~/.proprietary/derp/bin' and '~/.proprietary/bin' would be added,
+  #          but '~/.proprietary/foo/bar/bin' would not.
+  PATH="$(find -L $HOME/.proprietary -maxdepth 2 -mindepth 1 -name 'bin' -type d | tr '\n' ':')$PATH"
+fi
 
 # modify the prompt to contain git branch name if applicable
 git_prompt_info() {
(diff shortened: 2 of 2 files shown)


mike-burns:
Thanks for removing the `export` on `$PATH`. That's bugged me for a while.

I solve this problem by having multiple dotfiles repos, each of which has its own `bin` directory with scripts inside of it. Like:

```sh
$ ls bin
n -> /home/mike/.dotfiles-personal/bin/n
tat -> /home/mike/.dotfiles-thoughtbot/bin/tat
```

Given that, my rcrc(5) file would look like:

```sh
DOTFILES_DIRS="/home/mike/.dotfiles-personal /home/mike/.dotfiles-thoughtbot"
```

How do you feel about that approach? Would you still rather have them in a separate dotted directory?

TheLonelyGhost:
I have to admin I haven't really played around with multiple dotfile directories yet. It seems like it might require a certain structure to each of the named directories for `$DOTFILE_DIRS`, likely a format `rcm` is able to handle, is that right?

There are 2 reasons why I feel my solution might be a better fit overall:

1. Largely agnostic to the overall format of the repository (`bin/` containing executables is a reasonably well-accepted standard)
2. Even if `$HOME` directory is a soup of config directories and files, this groups project-specific helper repositories together under one directory, organizing scripts with similar roles in one's process. It could be project-based, software-based, or some other taxonomy that this change doesn't need to know about or care.

To give an example of how I currently use this, I group together some SVN wrapping scripts (to more easily enforce our workflow and provide help for those mainly familiar with git) that rely on a shared library of shell functions. Rather than have all of those bin files in my dotfiles's `bin/` directory, I made them their own repository. I also have a repository for deployment-related wrapping scripts (deploy to prod, integration, and specific branch testing server) and server triage scripts (reload db backup, restart all production servers on rotation, etc.).

All of our developers use SVN. That's not a problem to distribute (Group A). A subset of those have access and responsibility to run deployments (Group B). Another subset of the developers have access and responsibilities to do triage-related tasks on production (Group C). Group B and C might have overlap in members. I can grant access to these 3 repositories to members of each group depending on their role.

This is assuming a one-application team is using this. Another option is to have all those scripts lumped together in project-specific repositories, which might be removed from `$PATH` as clients are added and dropped.

TheLonelyGhost:
Rebased from upstream just now, though GitHub doesn't appear to note that on the PR. Mainly to incorporate upstream changes to handling PATH in `zshrc` instead, but also to address merge conflicts.

mike-burns:
> Rather than have all of those bin files in my dotfiles's bin/ directory, I made them their own repository.

a) Let's say that repo is checked out in a directory named `svn_dotfiles`. And let's say it looks like this:

```
svn_dotfiles
+-- bin
       +-- svn-git-1
        \-- svn-git-2
```
(That is, `~/svn_dotfiles/bin/svn-git-1` and `~/svn_dotfiles/bin/svn-git-2`.) Your `~/.rcrc` would look like this:

```sh
DOTFILES_DIRS="$HOME/svn_dotfiles $HOME/dotfiles"
```

When you run `rcup`, it will link both -- first your `svn_dotfiles` programs, and then your `dotfiles` ones.

If you wish to stop using the `svn_dotfiles`, you can run `rcdn -d ~/svn_dotfiles`.

b) If you instead want all programs in the same dotfiles repo, you can use a tag. Given the `bin/svn-git-1` program mentioned above, run `mkrc -t svn bin/svn-git-1` to tag it. (If you wish to do this manually, from your dotfiles repo do `mkdir -p tag-svn/bin`, and then move the program under `tag-svn/bin`.)

For more, see rcm(7).

Does this solve your use case?

TheLonelyGhost:
It sounds like the fact that it's a "dotfiles repository" doesn't mean it needs a particular structure, unless you want to use advanced features with `rcm` like tagging. It also appears that end users can add/remove scripts from their PATH without having to clone or delete any files, just run `rcdn`.

Here's the layout I have and the symlinks if I used your proposed method:

```
~/.proprietary
├── cape-deploy
│   ├── README.md
│   ├── bin
│   │   ├── deploy_cape_production  # -> ~/bin/deploy_cape_production
│   │   ├── deploy_cape_qa          # -> ~/bin/deploy_cape_qa
│   │   └── deploy_cape_staging     # -> ~/bin/deploy_cape_staging
│   └── cape-helpers.sh             # (Library)
├── cape-development
│   ├── README.md
│   └── bin
│       ├── cape_test          # -> ~/bin/cape_test
│       ├── load_sql_to_dev    # -> ~/bin/load_sql_to_dev
│       ├── sync_ci            # -> ~/bin/sync_ci
│       └── tag_rails          # -> ~/bin/tag_rails
├── cape-ops
│   ├── README.md
│   └── bin
│       ├── bounce-memory-production  # -> ~/bin/bounce-memory-production
│       ├── bounce-memory-qa          # -> ~/bin/bounce-memory-qa
│       └── check-memory-production   # -> ~/bin/check-memory-production
└── svn
    ├── README.md
    ├── bin
    │   ├── merge_branch    # -> ~/bin/merge_branch
    │   ├── new_branch      # -> ~/bin/new_branch
    │   ├── remove_branch   # -> ~/bin/remove_branch
    │   ├── switch_to       # -> ~/bin/switch_to
    └── svn-helpers.sh      # (Library)
```

The problem is `merge_branch` (found in `~/bin/`, symlinked to `~/.proprietary/svn/`) has a dependency on `../svn-helpers.sh`. Since we're calling the symlink, it will look for `~/svn-helpers.sh` instead of `~/.proprietary/svn/svn-helpers.sh`.

We could address this by symlinking `svn-helpers.sh` to the home directory as well, but then each repository like `svn/` would need to do that. It would add a lot of clutter very quickly.

We could also rewrite `merge_branch` to depend on it as an absolute path instead of a relative one, but then it might break as people have different names for the cloned dotfile repository directories.

My proposed method doesn't care about that at all. It keeps library code where it belongs: in the repository. Scripts can load a relative path to the library without problem. Adding a new repository is as simple as `git clone ...` to clone it into a subdirectory of `~/.proprietary/`.

mike-burns:
> The problem is merge_branch (found in ~/bin/, symlinked to ~/.proprietary/svn/) has a dependency on ../svn-helpers.sh

That is indeed a problem. You can fix this in an impressively complex way -- again, rcm has an example: https://github.com/thoughtbot/rcm/blob/master/bin/lsrc.in#L3-L4 -- or add e.g. `~/.local/share` and symlink it into there, and source it as `$HOME/.local/share`.

> It keeps library code where it belongs: in the repository. 

We seem to fundamentally disagree on what the dotfiles repo is used for. @geoffharcourt , as dotfiles maintainer, do you have any thoughts here?

geoffharcourt:
As it stands, I think our team is currently well-served by `rcm` in these situations. I'm reluctant to merge changes in that I don't believe members of our team would use (I'm currently trying to catalog what parts of the setup are in active use and which are not so that we can trim further). Since this functionality can be incorporated solely through the use of `.local` zsh configs without changes upstream, I think that's the right place for it.

If you want to do a separate PR, we can bring in the change to remove `export` on `PATH`, or I can just crib that in a separate issue.

TheLonelyGhost:
Sure, I'll just open a separate PR for the PATH export part.

TheLonelyGhost:
Opened #441 for PATH export component.

PR #435, opened 2015-10-21
Allow project-specific vimrc

geoffharcourt:
@croaky made this change in one of our projects to allow spelling to be
conditionally turned on in a text-heavy project.

The setting is turned on after all other configuration is loaded,
including plugins, and `secure` is set so that a project-level `vimrc`
cannot run unsafe commands.

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -158,3 +158,9 @@ set diffopt+=vertical
 if filereadable($HOME . "/.vimrc.local")
   source ~/.vimrc.local
 endif
+
+" Safely allow project-specific vimrc (check for `.git/safe`)
+if isdirectory(getcwd() . ".git/safe")
+  set secure
+  set exrc
+endif


vimrc
```diff
@@ -158,3 +158,6 @@ set diffopt+=vertical
 if filereadable($HOME . "/.vimrc.local")
   source ~/.vimrc.local
 endif
+
+set secure
```
teoljungberg: What do you think of setting these settings only when `.git/safe` is in place in the repository?

I imagine it were to looks something like:

```
if isdirectory(getcwd() . ".git/safe")
  set secure
  set exrc
endif
```

That way - we can be sure that the users vim configuration isn't going to be compromised.

Thoughts?

geoffharcourt: I like that a lot. Should we use Git to check further up the tree so that the settings work if you're not at the root folder, or is that getting too cute with configuration magic?

teoljungberg: It might make sense to add in the future, but I think that it might be overkill for the initial implementation.
I would say that we can add it when we feel like it's an issue

geoffharcourt:
The only concern I have at the moment is that there's no way for a user to disable this functionality with local overrides.

mike-burns:
This does not seem like a thing we'd want to default on.

> allow spelling to be conditionally turned on

If we merge this, then how do we conditionally turn spelling on?

teoljungberg:
@geoffharcourt They would be able to if we move the newly added paragraph above the sourcing of `.vimrc.local`. Then the users can permanently opt out by setting:

```viml
set secure&
set exrc&
```

In their `vimrc.local`. This would set `secure` and `exrc` back to their default values.

vimrc
```diff
@@ -158,3 +158,9 @@ set diffopt+=vertical
 if filereadable($HOME . "/.vimrc.local")
   source ~/.vimrc.local
 endif
+
+" Safely allow project-specific vimrc (check for `.git/safe`)
```
mike-burns: Typo fix:

```vim
" Allow project-specific vimrc (check for `.git/safe`)
```

geoffharcourt: Point taken.

geoffharcourt:
I think this was a little aggressive for shared files. I'm going to close this down.

PR #437, opened 2015-10-25
Bring shell config into (better) shellcheck compliance

geoffharcourt:
While evaluating linting and testing options for our shell
configuration, a run of Shellcheck against our executables turned up
some places where we could use more consistent syntax across our
executables, such as always using a shebang or quoting `$HOME` when we
build up a larger directory.

### DIFF
--- a/bin/git-ca
+++ b/bin/git-ca
@@ -1,1 +1,3 @@
-git commit --amend -v --date=$(date +%Y-%m-%dT%H:%M:%S)
+#!/bin/sh
+
+git commit --amend -v --date="$(date +%Y-%m-%dT%H:%M:%S)"
--- a/bin/replace
+++ b/bin/replace
@@ -9,4 +9,4 @@ shift
 replace_with="$1"
 shift
 
-ag -l --nocolor "$find_this" $* | xargs sed -i '' "s/$find_this/$replace_with/g"
+ag -l --nocolor "$find_this" "$@" | xargs sed -i '' "s/$find_this/$replace_with/g"
--- a/git_template/hooks/ctags
+++ b/git_template/hooks/ctags
@@ -3,7 +3,7 @@
 set -e
 
 PATH="/usr/local/bin:$PATH"
-dir="`git rev-parse --git-dir`"
+dir="$(git rev-parse --git-dir)"
 trap 'rm -f "$dir/$$.tags"' EXIT
 git ls-files | \
   "${CTAGS:-ctags}" --tag-relative -L - -f"$dir/$$.tags" --languages=-javascript,sql
--- a/zsh/functions/g
+++ b/zsh/functions/g
@@ -1,8 +1,8 @@
 # No arguments: `git status`
 # With arguments: acts like `git`
 g() {
-  if [[ $# > 0 ]]; then
-    git $@
+  if [[ $# -gt 0 ]]; then
+    git "$@"
   else
     git status
   fi
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -1,12 +1,12 @@
 #!/bin/bash
 
-touch $HOME/.psqlrc.local
+touch "$HOME"/.psqlrc.local
 
-if [ ! -e $HOME/.vim/autoload/plug.vim ]; then
-  curl -fLo $HOME/.vim/autoload/plug.vim --create-dirs \
+if [ ! -e "$HOME"/.vim/autoload/plug.vim ]; then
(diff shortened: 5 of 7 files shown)


geoffharcourt:
cc @thoughtbot/shell

bin/bundler-search
```diff
@@ -12,4 +12,5 @@
 #  2. Which gem names to search (defaults to all gems)
 
 pattern="$1"; shift
+# shellcheck disable=SC2046
```
gabebw: What is this ignoring?

I think it's worse to have this comment than to have shellcheck fail, because this comment is totally opaque.

When/if we run shellcheck against this again, I think we can make a judgment call.

zsh/functions/g
```diff
@@ -1,8 +1,8 @@
 # No arguments: `git status`
 # With arguments: acts like `git`
 g() {
-  if [[ $# > 0 ]]; then
```
gabebw: I actually prefer `>`, and I believe it's exactly the same as `-gt` in zsh.

gabebw: I was wrong: http://zsh.sourceforge.net/Doc/Release/Conditional-Expressions.html

`>` is ASCII comparison ("z" > "a") and `-gt` is numeric comparison.

bin/bundler-search
geoffharcourt: Shellcheck dies if you run it against multiple files and there's a failure earlier on.

I'm working on a test suite that's going to run our dotfile aliases, programs, and functions on different platforms to test, and was planning for shellcheck to be the linting portion of this work. Maybe we should look at doing shellcheck in Hound?

zsh/functions/g
geoffharcourt: Yeah, this was a big learning moment for me.

bin/replace
```diff
@@ -9,4 +9,4 @@ shift
 replace_with="$1"
 shift
 
-ag -l --nocolor "$find_this" $* | xargs sed -i '' "s/$find_this/$replace_with/g"
+ag -l --nocolor "$find_this" "$@" | xargs sed -i '' "s/$find_this/$replace_with/g"
```
geoffharcourt: @gabebw This and the corresponding quoting in `g()` were the two changes I was most concerned about. I was unable to break them in my personal testing, but wasn't sure if there's nuance I'm missing.

gabebw: There are two things happening here, `$*` vs `$@` and quoting.

Here's a shell script that will illustrate the difference:

```sh
#!/bin/sh

for argument in $*; do
  echo "$argument"
done
```

(For each bit about `$*` vs `$@` vs `"$@"`, assume I changed the `for` loop in the script accordingly.)

`$*`, with no quotes, will totally ignore spaces in arguments, even if the arguments are quoted:

```sh
# $*
$ ./demo.sh "keep it together" two
keep
it
together
too
```

Now let's try it with `"$*"`:

```sh
# "$*"
$ ./demo.sh "keep it together" two
keep it together too
```

As you can see, it interpreted all arguments as one huge string, instead of `keep it together` and `two` as separate strings. (So, it's wrong.)

Now let's try with `$@`:

```sh
# $@
$ ./demo.sh "keep it together" two
keep
it
together
too
```

It acts just like `$*`.

And finally, the one that we almost always want to use, `"$@"`:

```sh
# "$@"
$ ./demo.sh "keep it together" two
keep it together
too
```

At last, we have printed two lines for two arguments, appropriately split up according to how we passed them in as arguments.

gabebw: tl;dr always use `"$@"`

geoffharcourt: :+1: this is awesome. Thanks!

bin/git-ca
```diff
@@ -1 +1,2 @@
-git commit --amend -v --date=$(date +%Y-%m-%dT%H:%M:%S)
+#!/bin/sh
```
gabebw: I'd add a newline after this, for style.

gabebw:
One comment then :+1:

geoffharcourt:
@gabebw thanks for the review and the info.

git_template/hooks/ctags
```diff
@@ -3,7 +3,7 @@
 set -e
 
 PATH="/usr/local/bin:$PATH"
-dir="`git rev-parse --git-dir`"
+dir="$(git rev-parse --git-dir)"
```
mike-burns: Take that, Solaris!

PR #439, opened 2015-10-26
Replace the g function with gitsh

mike-burns:
A couple of years back, thoughtbot wrote gitsh. It's a powerful shell
for interacting with git, but despite its power it is extremely simple
to use. Running the empty command will run a (configurable) `git
status`, but running subcommands (e.g. `commit -v`) will act like `git`
was prepended. The `git` program itself can be replaced, for use with
`hub` for example.

This deprecates this `g` function.


calleerlandsson:
I've been using gitsh for almost two years now and am very happy with it.

christoomey:
I'm a recent but very happy convert to the gitsh way of life, and have recently been considering this move for myself as I still have some muscle memory around this `g` shortcut.

That said, I'm not sure removing this is the best option for the thoughtbot dotfiles. I think the `g` function is still a good choice for many users and has some benefits, primarily in the scope of tab completion, that gitsh does not have.

derekprior:
> That said, I'm not sure removing this is the best option for the thoughtbot dotfiles. I think the g function is still a good choice for many users and has some benefits, primarily in the scope of tab completion, that gitsh does not have.

What if we replaced it with an alias?

croaky:
I use the `g` function every day, haven't made `gitsh` part of my workflow yet. Happy to move this function to my personal dotfiles if it gets removed, but I'm guessing it's useful for a lot of people who use `thoughtbot/dotifles`.

geoffharcourt:
You'll have to pry `g()` from my cold, dead hands.

I'm giving `gitsh` a try and it's an interesting approach to git, but I use `g` to check for status even when I'm not getting ready to make a commit or perform other operations.

MattMSumner:
`g` is a big part of my workflow. If this was removed I would add it to my personal dotfiles.

blainesch:
:+1: I :heart: gitsh!

mikegee:
The title of this PR implies `g` will do something new with `gitsh`, but it just removes it. :disappointed: 

I'm interested in using `gitsh`, but `g` is in my muscle memory now.

geoffharcourt:
I think there are enough folks who are very heavily leaning on `g` that it doesn't make sense to replace it. I'm going to close this PR and leave our precious `g` function in place.

PR #440, opened 2015-10-26
PATH cleanup in zshrc

TheLonelyGhost:
There has been an explosion in my `$PATH` whenever I open a new tab in tmux, specifically with duplicate path entries.

```
# Before:
echo $PATH
# => "/Users/alexanderdav/.proprietary/cape-deploy/bin:/Users/alexanderdav/.proprietary/cape-development/bin:/Users/alexanderdav/.proprietary/cape-ops/bin:/Users/alexanderdav/.proprietary/svn/bin::.git/safe/../../bin:/Users/alexanderdav/.rbenv/shims:/Users/alexanderdav/.bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/Users/alexanderdav/.tmuxifier/libexec:.git/safe/../../bin:/Users/alexanderdav/.rbenv/shims:/Users/alexanderdav/.proprietary/cape/bin:/Users/alexanderdav/.rbenv/bin:/Users/alexanderdav/.tmuxifier/bin:/Users/alexanderdav/.bin:/usr/local/sbin"

# After:
echo $PATH
# => "/Users/alexanderdav/.proprietary/cape-deploy/bin:/Users/alexanderdav/.proprietary/cape-development/bin:/Users/alexanderdav/.proprietary/cape-ops/bin:/Users/alexanderdav/.proprietary/svn/bin:.git/safe/../../bin:/Users/alexanderdav/.rbenv/shims:/Users/alexanderdav/.bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/Users/alexanderdav/.tmuxifier/libexec:/Users/alexanderdav/.proprietary/cape/bin:/Users/alexanderdav/.rbenv/bin:/Users/alexanderdav/.tmuxifier/bin"
```

See commit messages for specific changes, but overall this does the following:

- PATH helper functions (for use within config files, primarily)
- Removing duplicate PATH entries
- Cleaning up empty entries in PATH

### DIFF
--- a/zsh/functions/move_to_front_of_path
+++ b/zsh/functions/move_to_front_of_path
@@ -1,0 +1,10 @@
+# Reorders existing entry in PATH to have it at the front
+#
+#   move_to_front_of_path "$HOME/.bin"
+#
+move_to_front_of_path() {
+  if [[ ":$PATH:" == *":$1:"* ]]; then
+    PATH=$(echo $PATH | sed 's#'$1'##g' | sed s/:://g | sed s/:$//g | sed s/^://g)
+    PATH="$1${PATH:+":$PATH"}"
+  fi
+}
--- a/zshrc
+++ b/zshrc
@@ -3,7 +3,7 @@ export VISUAL=vim
 export EDITOR=$VISUAL
 
 # ensure dotfiles bin directory is loaded first
-export PATH="$HOME/.bin:/usr/local/sbin:$PATH"
+path=($HOME/.bin /usr/local/sbin $path)
 
 # load rbenv if available
 if command -v rbenv >/dev/null; then
@@ -113,3 +113,16 @@ _load_settings "$HOME/.zsh/configs"
 
 # Local config
 [[ -f ~/.zshrc.local ]] && source ~/.zshrc.local
+
+# mkdir .git/safe in the root of repositories you trust
+path=(.git/safe/../../bin $path)
+
+# readjust order of PATH entries
+move_to_front_of_path "$HOME/.bin"
+move_to_front_of_path ".git/safe/../../bin"
+
+# Remove dupes from PATH
+typeset -U path
+
+# Solidify PATH changes for the rest of the shell sessions
+export PATH


srstevenson:
@TheLonelyGhost you could simplify this using zsh's `path` variable. `PATH` is actually a scalar whose contents are those of `path` array concatenated together with colons: see "PARAMETERS USED BY THE SHELL" in [zshparam(1)](http://linux.die.net/man/1/zshparam) for the details.

You can remove duplicate entries from `path`, keeping only the first occurence, with `typeset -U path` (I remember this mnemonically as U for unique). Appending and prepending to the path without duplication is then simply `path=($path /foo/bar)` and `path=(/foo/bar $path)` respectively. You can use the `NULL_GLOB` glob qualifier to add an entry only if the directory exists: `path=(/foo/bar(N) $path)` ([zshexpn(1)](http://linux.die.net/man/1/zshexpn) documents this and other qualifiers).

TheLonelyGhost:
Huh, that's a much better plan than the one I suggested. I've rewritten and force-pushed since it covers the same domain.

geoffharcourt:
If you reorder some of the path adjustments you're making in `zshrc` that you can avoid having to create that function that reorders the path at all.

geoffharcourt:
When I use this branch of dotfiles, I get an invalid path:

```
.git/safe/../../bin:/Users/geoff/.bin:/Users/geoff/.rbenv/shims/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
```

There's no `:` separating the rbenv shims from `/usr/local/sbin`, which breaks rbenv.

geoffharcourt:
The broken paths appear to be due to the `move_to_front_of_path` function. When I comment the use of that function out I get a non-rbenv-breaking `PATH`.

Beyond cleaning up the bug in the function, I'm extremely reluctant to introduce a function with this purpose into dotfiles when we can handle the quirks of path setup within `zshrc` through ordering. I think using `typeset -U path` to clean up the path is a great addition (and resolves the similar report in #443), and using the `path` variable looks like a cleaner way to write out path changes, but I don't want path setup to be dependent on a function that might not be available if a user hasn't run `rcup` yet.

If you can accomplish the path cleanup without an additional function I think this is a good candidate to merge.

geoffharcourt:
As @mike-burns noted in #441, you can remove `export PATH` here as well.

geoffharcourt:
@TheLonelyGhost what do you think about removing the path-editing function but retaining the `typeset` command to deduplicate paths?

geoffharcourt:
Rebased with just the `PATH` de-duplication without the additional function in f7c73f7. Thank you very much for your contribution!

PR #441, opened 2015-10-27
Defers 'export' until the end of file

TheLonelyGhost:
According to http://stackoverflow.com/a/7411509 the only reason to use
'export' is if you want to make variables set in the shell available
outside of the current shell or subshell. There's no obvious reason why
this is happening multiple times so it should only happen once at the
end of the configuration file.

Is there some reason why we're exporting each variable upon
instantiation?

### DIFF
--- a/zshrc
+++ b/zshrc
@@ -1,9 +1,9 @@
 # use vim as the visual editor
-export VISUAL=vim
-export EDITOR=$VISUAL
+VISUAL=vim
+EDITOR=$VISUAL
 
 # ensure dotfiles bin directory is loaded first
-export PATH="$HOME/.bin:/usr/local/sbin:$PATH"
+PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
 # load rbenv if available
 if command -v rbenv >/dev/null; then
@@ -11,7 +11,7 @@ if command -v rbenv >/dev/null; then
 fi
 
 # mkdir .git/safe in the root of repositories you trust
-export PATH=".git/safe/../../bin:$PATH"
+PATH=".git/safe/../../bin:$PATH"
 
 # modify the prompt to contain git branch name if applicable
 git_prompt_info() {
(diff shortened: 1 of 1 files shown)


geoffharcourt:
cc: @thoughtbot/shell

mike-burns:
Do PATH and PS1 need to be exported? They're only used by the shell.

gabebw:
I believe PATH does need to be exported but I don't remember why.

geoffharcourt:
From my research it looks as if it's [actually bad to export `PATH` in POSIX shells](http://unix.stackexchange.com/a/26059), but only due to `export`'s interpretation of `~` as a literal and not a stand-in for `$HOME`. (Since we prefer the more portable `$HOME` anyway, this looks harmless in practice.) I'm in favor of not `export`-ing things we don't need to export, so cutting out exports of `PATH` and `PS1` as suggested above looks good to me.

I am less in favor of declaring all the exports at the bottom of `zshrc` for a couple of reasons. First, I think [we're about to split of `zsh` configuration into multiple files](https://github.com/thoughtbot/dotfiles/pull/430), at which point we'll be declaring our variables in a greater number of separate locations anyway. Second, declaring all exports in one place implies to the reader that this particular location is where all exports occur, but users might export in other locations such as `zshrc.local`. 

Since an exported environment variable will track mutations made without the need to be re-`export`-ed, I think declaring intent close to the location the variable is declared will be easier to read.

jferris:
> Do PATH and PS1 need to be exported? They're only used by the shell.

Will Vim commands see programs from your `PATH` if it isn't exported?

jferris:
> Will Vim commands see programs from your `PATH` if it isn't exported?

It seems like (by default) `PATH` entries set in `.zshenv` are visible to Vim, but not `.zshrc`, because Vim runs in a non-interactive shell:

> On Unix the command normally runs in a non-interactive shell. If you want an interactive shell to be used (to use aliases) set 'shellcmdflag' to "-ic".

geoffharcourt:
@jferris, when I remove the `export` statements on `PATH` and then launch a new shell, open vim, and execute `:! echo $PATH` I see the full path. I think this behavior is due to rbenv `export`ing `PATH` [here](https://github.com/sstephenson/rbenv/blob/master/libexec/rbenv-init#L82). In this case, it looks as if debating whether to stop `export`ing `PATH is a moot point unless there's a change in rbenv.

geoffharcourt:
This of course does not apply in MacVim, where the `PATH` already isn't picking up our changes from `zshrc`.

geoffharcourt:
I think after 31b6ad4f3fd70732ae16b7093b4e4b8705b4a608 this change makes less sense, because we're declaring variables in separate files and it would be confusing to export them in a different file than that in which the variable was declared.

I've opened #446 to not export `PATH` or `PS1` at @mike-burns suggestion.

@TheLonelyGhost thank you for your contribution, I'm sorry this work was made less relevant by other changes that were recently made.

PR #442, opened 2015-10-29
Ignore `.env.local` instead of `.env`.

tute:
It makes sense to version keys used in local development in the `.env`
file, rather than constantly copying them from `.sample.env`.  You can
still override these locally with the `.env.local` file.

- Removes `.env` from the `gitignore` file.
- Adds `.env.local` to the `gitignore` file.

See:

* https://github.com/bkeepers/dotenv#should-i-commit-my-env-file
* https://github.com/thoughtbot/dotfiles/pull/369
* https://github.com/thoughtbot/suspenders/pull/563

### DIFF
--- a/gitignore
+++ b/gitignore
@@ -1,7 +1,7 @@
 .DS_Store
 *.sw[nop]
 .bundle
-.env
+.env.local
 db/*.sqlite3
 log/*.log
 rerun.txt


teoljungberg:
What do you think about adding support to foreman to load `.env.local` into suspenders to go along this change?

We solved it like [this] in Hound

[this]: https://github.com/thoughtbot/hound/commit/c4255796ff8e5ae122b81ee2e2e6d54da5b2126f

geoffharcourt:
This shouldn't be an overriding concern, but Forego (a fast Foreman implemented in Go) doesn't work with `.env.local` at the moment. I'm really liking Forego, so I want to hold this item for more feedback. I think we should definitely ignore `.env.local`, I'm just not sure if we want to add `.env` in since it might be needed for some projects that don't recognize the local file and clients may only be using `.env` and expect it to be ignored.

calebthompson:
This scares me a bit. I don't want unreviewed changes going into my environment, in case something is accidentally changed in a way I don't want my system to operate.

geoffharcourt:
@calebthompson The flip side to that would be users accidentally checking in API credentials. I think your concern is valid, but there is a cost to not ignoring it.

calebthompson:
I'd love to see anything dotenv loads by default be ignored. `.env*` basically.

calebthompson:
I'd say it's much more likely that we'll see API creds in a `.env` file than a `.env.local`, since we've trained ourselves to use it that way.

tute:
> @calebthompson says: I'd say it's much more likely that we'll see API creds in a .env file than a .env.local, since we've trained ourselves to use it that way.

> @thorncp says (https://github.com/thoughtbot/suspenders/pull/563#issuecomment-152219147): A common pattern I've seen is putting placeholder variables for sensitive info in .env, then overriding them in .env.local. The placeholder signifies what variables the app expects to be there. Here's an example from Hound.

I don't find the need to use non staging services secret keys in development. Do we sometimes encounter that need? If so, should they connect to staging accounts? If not possible, we'll catch the API creds in code review, and they won't be commited.

geoffharcourt:
I'm leaning toward @calebthompson's suggestion of just ignoring anything `dotenv` takes in with a globbing ignore statement.

geoffharcourt:
@tute, what do you think about ignoring _both_ `.env` and `.env.local`?

tute:
> What do you think about ignoring both .env and .env.local?

I don't want to ignore `.env` for the reasons outlined in https://github.com/bkeepers/dotenv#should-i-commit-my-env-file:

> Personally, I prefer to commit the .env file with development-only settings. This makes it easy for other developers to get started on the project without compromising credentials for other environments. If you follow this advice, make sure that all the credentials for your development environment are different from your other deployments and that the development credentials do not have access to any confidential data.

This has proven useful in a few projects I worked at or were people I know have worked at, so you don't need to remember to copy `.sample.env` or it's variations, or fix broken apps for confused team members through Slack or ScreenHero.

geoffharcourt:
I think we're at an impasse here over approaches. 

Getting `.env` into source control allows us to follow a "batteries included" philosophy where non-confidential development or dummy credentials can be pre-supplied and kept in sync on applications. However, doing so means that changes to your system that you haven't reviewed can come in upstream from source control (whether malicious or otherwise) that might have adverse impact on your local environment.

Ignoring everything is probably better than what we're doing now (only ignoring `.env` but not ignoring `.env.local` is problematic), but doesn't solve the originally reported problem.

I'm leaning toward @tute's solution because sharing an application via source control already entails the risk of unreviewed changes impacting your application's operation (the changes could be in the code and even harder to spot than in the `.env` file), but I'd like to hear further feedback before we take action.

mike-burns:
`bin/setup` copies `.sample.env` into `.env`, typically.

derekprior:
> doing so means that changes to your system that you haven't reviewed can come in upstream from source control (whether malicious or otherwise) that might have adverse impact on your local environment.

This is true of any code you pull from anywhere and execute, regardless of file name. I'm not sure this is worth worrying about for this change.

I don't think there's a perfect solution. The `.sample.env` file has worked okay. It's a bit confusing when you need to update the `.env` file because just re-running a (hopefully idempotent) `bin/setup` file might clobber some keys I pulled down from staging.

I've seen a few recent projects where `bin/setup` went ahead and pulled down any truly "secret" configuration keys from the heroku staging environment. That was really pretty slick and meant that running `bin/setup` was the way to get your environment right and you didn't lose anything.

jferris:
The most common problem I have with `.sample.env` is that running `bin/setup` doesn't add new keys. That means you can start getting `fetch` or `nil.NoMethodError` issues when doing a `git pull`.

Improving `bin/setup` to somehow merge new keys from `.sample.env` into `.env` would be neat-o.

geoffharcourt:
I'd be worried about `bin/setup` bringing in keys that might start using the network in development. I often populate `.env` with hostnames that prevent workers in development from hitting the web (WebMock would block these things in testing, but not in development environments).

Agreed that the current `.sample.env` process isn't intuitive. The merging idea is really promising.

tute:
> Improving bin/setup to somehow merge new keys from `.sample.env` into `.env` would be neat-o.

If we merged this, there would be no `.sample.env` to merge from. What exactly do you mean?

tute:
Will move forward with these changes, which are simple to override for the people who would rather not use this workflow.

@geoffharcourt, suspenders uses `dorenv-rails`, and as outlined in https://github.com/bkeepers/dotenv#multiple-rails-environments, Teo's https://github.com/thoughtbot/dotfiles/pull/442#issuecomment-152122856 about how they do it in Hound works well.

Will merge suspenders and then we do this one!

tute:
Suspenders merged, this is ready to go! :)

geoffharcourt:
@lpender can you update the commit message with a sentence explaining that Foreman no longer supports Dotenv (and it's use of `.env.local`) as a reason for making this change? I can go ahead and merge it after that's updated.

tute:
@geoffharcourt I rebased the branch.

I'm not sure how to add the context about foreman you ask for. We don't use foreman anymore in suspenders (we use `heroku local` instead), and `.env.local` is still useful through `dotenv-rails` gem. Am I missing something?

gylaz:
@tute I believe the issue now is that `heroku local` does not pick up on `.env.local`. Since we don't use Foreman anymore, we can't take advantage of providing `env: .env.local` in `.foreman`. I'm not sure how we can point `heroku local` to pick up on `.env.local`.

lpender:
Hey guys, thanks for your comments.

If `heroku local` doesn't support `.env.local` maybe the solution is to add `dotenv` to the `gemfile`?

Let me do some testing and get back.

tute:
@schneems how can we make `heroku local` take into account the `.env.local` file (or similar)? Thanks!

calebthompson:
Probably better off asking someone involved in that project: https://github.com/heroku/heroku-local

tute:
> Probably better off asking someone involved in that project: https://github.com/heroku/heroku-local

Indeed. @dickeyxxx, how can we make `heroku local` take into account the `.env.local` file (or similar)? Thanks!

dickeyxxx:
You just specify `-e`. It's just a wrapper to forego.

gylaz:
@dickeyxxx We want both `.env` and `.env.local` to be considered, with keys from later overwriting keys from former.

dickeyxxx:
We can't simply add `.env.local` as a file to look for in forego, that would break people's setups if they had a `.env.local` they didn't want to use.

jferris:
Is there a way for us to source `.env.local` from `.env`?

teoljungberg:
@jferris I don't think so.

geoffharcourt:
If a key is specified twice in an .env file, does the final specification get priority?

We could concatenate the two files on server start and have forego look at the result.

geoffharcourt:
As `.env.local` support won't be available in forego, I don't think this PR can be merged. If forego adds support or if we find an alternative process server for local development, we can revisit.

PR #445, opened 2015-11-06
Add custom ruby installation function

graysonwright:
Based on discussion at https://trello.com/c/0rlBCEms/411-ruby-install
Co-dependent on https://github.com/thoughtbot/laptop/pull/419

## Problem:

The two main tools for installing ruby, `ruby-build` and `ruby-install`,
both have drawbacks.

- `ruby-build` requires an update in order to install new versions of
  ruby.
- `ruby-install` requires command-line arguments in order to install
  rubies in an rbenv-compatible way.

## Solution:

Add a wrapper around `ruby-install` that takes care of the command line
arguments for us. This lets us install new ruby versions with the
command:

```
rinstall 2.2.2
```

where 2.2.2 is the version number.

The tool never needs to be updated, and installs into the directory that
rbenv expects. As a bonus, it will automatically install bundler for you
and attempt to `bundle install` for the current directory.

### DIFF
--- a/zsh/functions/rinstall
+++ b/zsh/functions/rinstall
@@ -1,5 +1,0 @@
-function rinstall () {
-  ruby-install --install-dir "$HOME/.rbenv/versions/$1" ruby "$1"
-  gem install bundler --conservative
-  bundle
-}
--- a/zsh/functions/rinstall
+++ b/zsh/functions/rinstall
@@ -1,0 +1,5 @@
+function rinstall () {
+  ruby-install --install-dir "$HOME/.rbenv/versions/$1" ruby "$1"
+  gem install bundler --conservative
+  bundle
+}
--- a/bin/rinstall
+++ b/bin/rinstall
@@ -1,0 +1,11 @@
+!#/bin/sh
+
+if command -v ruby-install >/dev/null; then
+  ruby-install --install-dir "$(rbenv root)/versions/$1" ruby "$1" && \
+    gem install bundler --conservative
+
+  [ -f Gemfile ] && bundle
+else
+  echo "You need to install ruby-install first."
+  return 1
+fi
--- a/zsh/completion/_rinstall
+++ b/zsh/completion/_rinstall
@@ -1,0 +1,5 @@
+#compdef rinstall
+
+if command -v rbenv >/dev/null; then
+  compadd $(rbenv completions install)
+fi


geoffharcourt:
This is probably much less relevant given that we're not doing significant work on the 1.9.x series of Ruby, but what do you think about adding tab completion such as rbenv/ruby-build have to complete available versions?

blainesch:
> Based on discussion at https://trello.com/c/0rlBCEms/411-ruby-install

It might be difficult for non-thoughtboters to give any kind of feedback if half the conversation is already private?

graysonwright:
@blainesch good point.

I've opened up a separate discussion about making our research board publicly-viewable, but until that time here's an export of the conversation.

[rintall.pdf](https://github.com/thoughtbot/dotfiles/files/34785/rintall.pdf)

geoffharcourt:
@graysonwright I've made a few changes in response to some problems that I had during testing. Can you take a look and squash them in if you approve? 

I'd love to add tab-completion for known versions that `rbenv` has knowledge of, but I think that's something that could be done in a later PR.

zsh/functions/rinstall
```diff
@@ -0,0 +1,13 @@
+function rinstall () {
+if command -v ruby-install >/dev/null; then
```
teoljungberg: Should this be indented one level?

geoffharcourt: Good point! Fixed.

mike-burns:
Why is this a function?

zsh/functions/rinstall
```diff
@@ -0,0 +1,13 @@
+function rinstall () {
```
mike-burns: Don't use the `function` keyword.

zsh/functions/rinstall
```diff
@@ -0,0 +1,13 @@
...
+  if command -v ruby-install >/dev/null; then
+    ruby-install --install-dir "$(rbenv root)/versions/$1" ruby "$1" && \
+      gem install bundler --conservative
+
+    if [ -f Gemfile ]; then
+      bundle
+    fi
```
mike-burns: Could be shortened:

```sh
[ -f Gemfile ] && bundle
```

geoffharcourt:
I like this approach, but FWIW, my approach to this has been to install ruby-build separately (clone it from Github, not to install through Homebrew) and then to have an `update-rubies` script that updates that repository. I've found that the ruby-build project tends to get a commit merged into master almost immediately when a new version comes out (vs. waiting for ruby-build _and_ the Homebrew formula to get merged upstream).

If there's interest in that approach, this is my `update-rubies` script: https://github.com/geoffharcourt/dotfiles-local/blob/2f8c0380ffeebcf0cc63b6007e83920f112d1646/zsh/functions/update-rubies

geoffharcourt:
@mike-burns would your recommendation be to put this in`bin`?

mike-burns:
Yes. Putting it in `bin/` makes it accessible to anyone of any shell, including vim. Functions are only needed for advanced zsh magic.

geoffharcourt:
@mike-burns @graysonwright I've moved the logic out from a function and into a script in `bin`. zsh is much less finicky about setting up tab-completion for a script than a function, so the tab-completion was much simpler to write.

I think retaining the ability to complete known (rbenv-known) Ruby versions is a solid blend of the advantages of @graysonwright and my prior approaches to this issue.

Should we be calling this `rbinstall` so that it's more clear what it does?

graysonwright:
@geoffharcourt since it's my last day, do you mind if I entrust this PR to you?

I like the `bin` script idea, and tab completion would be cool. I'm not sure about having the extra `update-rubies` step, but that's a pretty minor thing overall.

geoffharcourt:
@graysonwright that sounds great. I plan on merging this later in the week under your authorship if you don't have any further feedback.

gabebw:
I use the [rbenv-default-gems](https://github.com/rbenv/rbenv-default-gems) plugin, which installs gems after every `rbenv install`. Here's mine: https://github.com/gabebw/dotfiles/blob/master/tag-ruby/default-gems

Maybe it's better to use the rbenv plugin?

jferris:
> Maybe it's better to use the rbenv plugin?

Does that solve the original issue?

> ruby-build requires an update in order to install new versions of ruby.

gabebw:
Ah, `rbenv install` uses `ruby-build` behind the scenes. This PR is recommending we use `ruby-install` instead. Never mind.

geoffharcourt:
This approach has the advantage of not needing to wait for the `ruby-build` folks to make their update. However, when MRI 2.2.4 came out this morning, it was available on `ruby-build` immediately.

If folks are using `ruby-build` as installed by rbenv or through Homebrew this script would be very useful. I've opted to install it as a standalone from rbenv, so I don't have to wait for the Homebrew formula to update. I'm not sure how many other folks are doing that.

geoffharcourt:
After playing with `ruby-install` I'm sold on it as a solution for installing Ruby. 

ruby-build is a great tool, but the best way to use it is to install it separately from your package manager so that it can be updated independently of the package manager's own versioning of ruby-build. I've used this approach to get access to the newest Rubies almost as soon as they come out. However, while I'm willing personally to engage in the work required to maintain this approach on my machines, I don't think it's a great approach to recommend more generally. 

`ruby-install`'s approach to listing the latest Ruby versions works really well, and I like how the program is able to be aware of new releases of Ruby. Being able to install a version just by knowing the number even if the local installation isn't yet aware of the version is a nice convenience.

The solution advocated in this PR worked for me in testing, but I think one approach we should attempt before merging would be to convince the ruby-install team to support an environment variable or rc file to declare a default installation directory. Then, rather than requiring dotfiles users to learn/know a custom command that they would be using very infrequently, we could recommend that Rubies be installed using  ruby-install, and it would do so in an rbenv-compatible way.

I'm going to file a PR for ruby-install to advocate for this approach. If it is successful we'll close the PR, but what we have here is a good fallback option if the suggestion isn't adopted.

geoffharcourt:
@graysonwright what do you think about re-packaging this as a rbenv plugin that a user can use in place of `ruby-build`? We could then install that plugin in Laptop. The plugin could have the same `rbenv-install` command that ruby-build installs, but it could do the appropriate version-naming that the script you created does here.

croaky:
I suppose this is the `ruby-build` alternative:

https://github.com/thoughtbot/dotfiles/pull/467

I've had it on my own machine for awhile.

calebthompson:
RVM is also a solution to your ruby installation woes.

graysonwright:
@geoffharcourt I think that's a great idea, but I don't currently have the bandwidth to look into that. /=

geoffharcourt:
@graysonwright I'm sorry this has been sitting outstanding for so long. I've been trying to figure out the best approach (at this point, which project out of dotfiles / laptop / a new rbenv plugin) works going forward. 

I agree that the problem you're solving is worth solving!

geoffharcourt:
After a lot of mulling, I don't think this is a good fit for dotfiles. I think I would use something like it if it were packaged as an rbenv plugin or something else through a package manager.

PR #446, opened 2015-11-06
Don't export unneeded variables

### LINKED ISSUE #270: New prompt doesn't work when running as root
I was experimenting with some unix stuff today and noticed that the new shell prompt introduced in https://github.com/thoughtbot/dotfiles/pull/239 doesn't work properly when running as root. Here's the output:

```
~ % sudo su
Password:
sh: git_prompt_info: command not found
%{[blue]%}%c%{%} %#
```

geoffharcourt:
`PS1` and `PATH` don't need to be exported in order to perform their
duties in the shell. `PS1` in particular causes breakage if subshells
are launched that don't have access to our `git_prompt_info` function.

Note that `PATH` continues to get `exported` due to `rbenv`'s `init`
subcommand exporting the variable.

Fix #270.

### DIFF
--- a/zsh/configs/post/path.zsh
+++ b/zsh/configs/post/path.zsh
@@ -1,5 +1,5 @@
 # ensure dotfiles bin directory is loaded first
-export PATH="$HOME/.bin:/usr/local/sbin:$PATH"
+PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
 # load rbenv if available
 if command -v rbenv >/dev/null; then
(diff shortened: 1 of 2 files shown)


geoffharcourt:
Made obsolete by b3a3c2e67d6e732fe29c396403477474deef3fc4, f074afeae3cd9075c0cca1078bc7dfdaa447bc9f, and f7c73f7c1bb9a902ee3752bd156c095e59469f46.

PR #447, opened 2015-11-09
Fix tab-completion on `g` function

geoffharcourt:
After our reorganization of zsh configuration into separate files,
tab-completion for the `g` git function stopped working. This change
moves the setup for tab completion back into `zshrc` and restores the
completion functionality for `g`.

### DIFF
--- a/zsh/configs/post/completion.zsh
+++ b/zsh/configs/post/completion.zsh
@@ -1,6 +1,0 @@
-# load our own completion functions
-fpath=(~/.zsh/completion /usr/local/share/zsh/site-functions $fpath)
-
-# completion
-autoload -U compinit
-compinit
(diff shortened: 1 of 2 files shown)


MattMSumner:
This LGTM.

geoffharcourt:
@MattMSumner one question I have is why this wasn't working before. My research of `compdef` and `compinit` hasn't revealed why setting completion directly from `zshrc` works any differently (for a function at least) than it would when we set up completion from a file sourced in `zshrc`.

cc @thoughtbot/shell

maletor:
This seems like a step backwards. Also, why after the local sourcing?

geoffharcourt:
@maletor as it stands with the base dotfiles, loading completions in `zsh/configs/post` doesn't set completions up with the `g` function in my environment.

Since completions wouldn't be used any earlier in the shell initialization process, I don't see a reason why we shouldn't load them as late as possible in case there are redefinitions of functions, etc. that might happen in local configuration and undo completion setup.

geoffharcourt:
I think this was a false alarm. I'm closing unless we get further reported problems.

PR #448, opened 2015-11-09
zsh: add editor files to zshenv

maletor:
The zshenv file seems like the canonical place to keep the editor config
variables. The change also allows for easier overriding for other
editors by adding the editor overrides to zshenv.local

### DIFF
--- a/zsh/configs/editor.zsh
+++ b/zsh/configs/editor.zsh
@@ -1,2 +1,0 @@
-export VISUAL=vim
-export EDITOR=$VISUAL
--- a/zshenv
+++ b/zshenv
@@ -1,3 +1,6 @@
+export VISUAL=vim
+export EDITOR=$VISUAL
+
 local _old_path="$PATH"
 
 # Local config


geoffharcourt:
@maletor, you can override those variables by putting your own editor config in your local dotfiles to be symlinked to `zsh/configs/editor.zsh`. RCM will prioritize your personal configuration over the group configuration with the standard `rcrc`.

maletor:
I know that is the case, but you think they are better sourced in `zshrc` over `zshenv`?

geoffharcourt:
We've recently had some debate and discussion about where to set various environment variables in #426 and #307. At the moment we're going with [this instruction in `zsh`'s sample `zshenv`](http://sourceforge.net/p/zsh/code/ci/master/tree/StartupFiles/zshenv), which says not to assign things in `zshenv` unless you have a very good reason to do so. As `zshrc` is sourced in all interactive shells, I'm pushing to keep the `EDITOR/VISUAL` setup here so that it's configured as similarly as possible to the other settings in our shared `zsh` config.

geoffharcourt:
@maletor if you find something in a script that breaks without `EDITOR` being set, we should definitely discuss further.

My current goal with dotfiles is to get the configuration set up in pieces that are as small and manageable as possible so that many users can pick and choose the pieces they want to use. #430 was an early step in that direction. If we identify pieces that don't work well being sourced in `zshrc` we can look into items that need to be loaded in `zshenv`, but I think we will not have any code beyond our path warning function in `zshenv` in the near future.

maletor:
Definitely agree. `rcm` makes it all too easy to override as well.

PR #449, opened 2015-11-13
Add golang configuration samples

calebthompson:
* `softtabstop=2`: As a side effect of setting how "large" a tab is in
 insert mode, Vim is able to delete tabs with the backspace key in
 Insert mode.
* `compiler go`: Necessary for plugins using the `make` command, such as
 vim-dispatch.
* `listchars=tab:\ \ `: Don't highlight tabs as "»·" as if they were
 problem characters in go files. Also recommended for *sh files and any
 others which which should also be using tabs.
* `fatih/vim-go`: The de-facto Golang Vim plugin.

### DIFF
--- a/vim/ftplugin/go.vim
+++ b/vim/ftplugin/go.vim
@@ -1,0 +1,6 @@
+let g:go_fmt_command = "goimports"
+
+setlocal softtabstop=2
+setlocal listchars=tab:\ \ ,trail:·,nbsp:·
+
+compiler go
--- a/vimrc.bundles
+++ b/vimrc.bundles
@@ -28,8 +28,9 @@ call plug#begin('~/.vim/bundle')
 
 " Define bundles via Github repos
 Plug 'christoomey/vim-run-interactive'
-Plug 'kchmck/vim-coffee-script'
 Plug 'ctrlpvim/ctrlp.vim'
+Plug 'fatih/vim-go'
+Plug 'kchmck/vim-coffee-script'
 Plug 'pbrisbin/vim-mkdir'
 Plug 'scrooloose/syntastic'
 Plug 'slim-template/vim-slim'


vimrc
```diff
@@ -57,16 +67,11 @@ augroup vimrcEx
...
 
-" Softtabs, 2 spaces
-set tabstop=2
-set shiftwidth=2
-set shiftround
-set expandtab
-
-" Display extra whitespace
-set list listchars=tab:»·,trail:·,nbsp:·
+  let g:go_fmt_command = "goimports"
```
jferris: These seem like they make sense in a go plugin. Maybe a good pull request to vim-go?

jferris:
We've generally resisted adding configuration for languages and platforms that aren't universal at thoughtbot. If we switched to rcm-style tags, this could make more sense, but I don't think Go is currently widespread enough that it makes sense to modify our base configuration for it.

vimrc
calebthompson: These are settings provided to customize `vim-go`. As far as I know everyone who's set up their personal versions of these has this setting.

vimrc
```diff
@@ -57,16 +67,11 @@ augroup vimrcEx
...
+  autocmd FileType go compiler go
+  autocmd FileType go setlocal listchars=tab:\ \ ,trail:·,nbsp:·
```
croaky: I do this slightly differently but don't know enough about the differences between approaches to understand the tradeoffs:

" Don't highlight tabs in Go files
autocmd BufRead,BufNewFile *.go setlocal nolist ft=go

calebthompson: I believe `vim-go` sets the `filetype` command for us. Strictly speaking, the trailing character highlighting might not be useful as it's stripped out on write, but I kept it in for parity with other filetypes.

Basically, this way we still get the trailing character highlighting:



Whereas with `nolist` I see:

croaky:
Looks like a pretty good set of minimal changes.

Not suggesting adding any of these but for anyone following along and curious about Go Vim config, I've also got:

" Automatically import packages on file save
let g:go_fmt_command = "goimports"

" Bind Leader r to run current Go file
au FileType go nmap <Leader>r <Plug>(go-run)

" Bind Leader t to run current Go test suite
au FileType go nmap <Leader>t <Plug>(go-test)

" Bind Leader s to list interfaces for the type under cursor
au FileType go nmap <Leader>s <Plug>(go-implements)

" Bind Leader i to list type info for the type under cursor
au FileType go nmap <Leader>i <Plug>(go-implements)

" Syntax highlight Functions, Methods, and Structs
let g:go_highlight_functions = 1
let g:go_highlight_methods = 1
let g:go_highlight_structs = 1

croaky:
If we were to add this to the dotfiles, would we also want zsh config?

# Go
export GOPATH=$HOME/go
export PATH=$PATH:$GOPATH/bin
export GOROOT="$(go env GOROOT)"
export PATH=$PATH:$GOROOT/bin

calebthompson:
@jferris Right, that makes sense if users consider this not to be more of a repository of useful snippets, which I know most people do.

I'd like to keep this open so @thoughtbot/go can agree on a setup, then I don't mind closing.

croaky:
I think a number of us also have some git hooks for Go: https://github.com/croaky/dotfiles/tree/master/git_template.local/hooks

calebthompson:
@croaky interesting. I have a subset of those:

```
source ~/.dotfiles/homebrew/path.zsh # contains export PATH="/usr/local/bin:/usr/local/sbin:$PATH", which I gives me `go env`
export GOPATH=`dirname $(go env GOROOT)`
export GOBIN=/usr/local/bin
```

calebthompson:
> " Syntax highlight Functions, Methods, and Structs
> let g:go_highlight_functions = 1
> let g:go_highlight_methods = 1
> let g:go_highlight_structs = 1

@croaky can you show me the difference with having these / not? I just tried setting them and couldn't see a difference.

geoffharcourt:
Is this an opportunity for us to start using RCM tags as @jferris suggested? 

I've started becoming conscious of things that slow down shell initialization. I suspect that the suggested zsh setup (getting the Go-related environment variables set) would run very quickly despite executing commands.

geoffharcourt:
@calebthompson another possibility since this is all vim settings is to turn this in to a Vim plugin called `go-ralph-go`. Let's not overlook this opportunity.

bernerdschaefer:
@croaky @calebthompson If Go is installed correctly, you shouldn't need to set `GOROOT` (or use `go env GOROOT`). My shell-related setup for Go is:

```
GOPATH=$HOME
```

Packages built by `go install` then get picked up by my existing `PATH=$HOME/bin:$PATH` config.

@calebthompson Setting `GOPATH=$GOROOT` seems like a possibly problematic choice -- that means your external libraries are mixed up with the runtime and standard libraries. It also means you'll have to re-fetch all dependencies when you upgrade Go versions. If you switch to something like @croaky's setup (`GOPATH=~/go`), you won't have those problems, and you also wouldn't need to change `GOBIN`.

bernerdschaefer:
@jferris I don't understand the resistance to adding support for new languages. Can you explain further?

jferris:
@bernerdschaefer we had an [earlier proposal](https://github.com/thoughtbot/dotfiles/pull/196) to add syntax highlighting for Go, and we decided not to merge it.

I can see some logic to this:

* Everything we add will slow things down and potentially break things, so each user should only have things that are potentially useful for their work. Developers (and designers) who aren't working on Go projects don't benefit from these changes.
* People already complain that the dotfiles are too large, and adding detailed configuration for new languages will make them much larger. If we merge Go support, it seems like we have no reason not to merge support for other languages we're trying, like Elixir, Haskell, and Clojure.

Part of our earlier decision was based on the fact that it was fixing a short-lived bug, so it wasn't worth having something in our dotfiles. We also brought up the possibility of using rcm to make this easier, but we still haven't broken up the dotfiles so that it would be easy to use rcm to selectively pull in Go support.

I think it would actually be better to break up the dotfiles to remove some of the functionality many of us don't use. For example, all of the Rails-related stuff is pretty heavy weight and isn't useful for our mobile team.

geoffharcourt:
With regard to keeping vim fast: vim-plug has a `{'for': filetype}` argument that we could use to not load expensive plugins until a file that uses them is loaded. This feature unfortunately does not work for `vim-rails`, as some of the magic is needed before any Ruby files get loaded (the navigation commands), but  for any plugin that's filetype-dependent it could be useful.

Here's an example from my [vimrc.bundles.local](https://github.com/geoffharcourt/dotfiles-local/blob/edae1a78c9ea8cce2ce4e6f40a8a7e026a3dee2e/vimrc.bundles.local#L27) where I don't load a markdown-associated plugin until I view a markdown file.

One small obstacle to loading plugins on the basis of rcm tags is that we'll have to move the vim-plug loading statements into separate files. That's not insurmountable, but it's not as clean as our current plugin setup.

geoffharcourt:
Maybe tag-based plugin setups could be declared in `vimrc.bundles.[filetype]`, and then we could source any file in the home directory that followed that pattern.

teoljungberg:
It sounds like we should create a vim plugin

bernerdschaefer:
@jferris I've always thought part of the appeal of `dotfiles` was that there's a good configuration for a tool *when you reach for it*.

> Everything we add will slow things down and potentially break things, so each user should only have things that are potentially useful for their work.

We should absolutely reject plugins and configurations which slow things down for non-users. But well-written plugins and configurations shouldn't have that problem.

For example, including `vim-go` adds an average of 280 microseconds (that's 0.00028 seconds) to vim startup when opening a non-Go file.

> If we merge Go support, it seems like we have no reason not to merge support for other languages we're trying, like Elixir, Haskell, and Clojure.

I think we should absolutely merge support for these tools, assuming there's enough consensus among the users and a willingness to maintain the configurations.

jferris:
@bernerdschaefer I think all of that makes sense. I don't want to avoid Vim/shell support for Go out of fear without trying it, and I also realize I'm making the classic "slippery slope" argument. I think we should leave this open for a few more days to allow for further feedback, but I'm not opposed to merging it.

vimrc
```diff
@@ -29,6 +29,16 @@ endif
 
 filetype plugin indent on
 
+" Softtabs, 2 spaces
+set tabstop=2
+set shiftwidth=2
+set shiftround
+set expandtab
+set softtabstop=2
```
mike-burns: The Python folks (most non-Ruby folks, it seems) have a larger sts and ts. Is it reasonable to set this to 2 globally/is there a better global number to set it to?

geoffharcourt: If these were an `autocmd` for Go with `setlocal`s, all these settings would be Go-only, correct?

teoljungberg: That's true, I suggest placing them under ".vim/ftplugin/go.vim" and use "setlocal" instead of "set" to have them locally done per go files

---
Teo Ljungberg

geoffharcourt:
I moved @calebthompson's suggested changes (outside of the plugin addition) into `ftplugin` in 443cf69782dc5f4870d3d021268ba9a95bc3ca08. If we like that approach, we can squash the commits and merge.

I'm not an expert Go developer, but in my testing all the proper settings were in place. I didn't see any difference whether I set the `compiler` setting or not, so if someone who uses Go more regularly could test I'd appreciate it.

`vim-go` adds a trivial amount of startup time for Vim (0.123 miliseconds for a non-Go file when starting Vim), so I'm not concerned about adding it to our list of plugins.

I'm comfortable merging after the Go users confirm the configuration is correct.

calebthompson:
As discussed in Slack the other day, compiler sets up errors and makeprg for you. These are, respectively, important and nice to have. If prefer to keep that setting.

geoffharcourt:
@calebthompson, I'm not proposing to remove the setting, I've just moved it to `ftplugin` so that it will be applied to Go files.

croaky:
> `export GOPATH=$HOME`

I like it! Just switched over to that in my personal dotfiles. Should we add that to https://github.com/thoughtbot/dotfiles/blob/master/zsh/configs/post/path.zsh?

croaky:
Testing out this branch, I found I still needed:

```
" Don't highlight tabs in Go files
autocmd BufRead,BufNewFile *.go setlocal nolist ft=go
```

In order to look like:

screen shot 2015-11-24 at 10 18 39 pm

Instead of:

screen shot 2015-11-24 at 10 19 02 pm

croaky:
```
let g:go_highlight_functions = 1
let g:go_highlight_methods = 1
let g:go_highlight_structs = 1
```

With:

screen shot 2015-11-24 at 10 26 57 pm

Without:

screen shot 2015-11-24 at 10 26 30 pm

geoffharcourt:
I made some changes to our Gitcommit, Markdown, and CSS/Sass autocommands in https://github.com/thoughtbot/dotfiles/pull/462, I think that could serve as a template for implementing buffer-specific Go language changes.

geoffharcourt:
With the config moved to `ftplugin/go.vim`, the compiler setting and other options were correctly set for Go files (running `:make` invoked Go). @calebthompson does this look OK? If you approve and want to squash/rebase under your name I can go ahead and merge this.

calebthompson:
Merged.

geoffharcourt:
Thanks @calebthompson!

croaky:
Whoa, awesome!

PR #453, opened 2015-11-16
Remove Unused `move_to_front_of_path` Function

reshleman:
In #440, a `move_to_front_of_path` function was introduced, but
was [ultimately not used][1] when these changes were merged as f7c73f7.
The function itself was not removed, however.

This commit removes the unused function.

[1]: https://github.com/thoughtbot/dotfiles/pull/440#issuecomment-156549420


geoffharcourt:
@reshleman this was my mistake, I'll get that removed now.

@blainesch I have a [PR for RCM](https://github.com/thoughtbot/rcm/pull/167) that I'm working on that might help, but it will not be finished very soon.

geoffharcourt:
Thanks @reshleman!

geoffharcourt:
@blainesch the resolution for removing vestigial code will be RCM functionality rather than something in dotfiles.

PR #454, opened 2015-11-16
Remove unused path-reordering function

geoffharcourt:
In f7c73f7 we introduced a suggested function to reorder the path. In
the process of rebasing that solution and shrinking its scope, the
function was not removed although it is not called from the `zsh`
configuration. Removing it to clean up the repository.

Close #453.


geoffharcourt:
Duplicate fix.

PR #455, opened 2015-11-18
Set `<leader>ct` shortcut to use Git ctags hook

geoffharcourt:
In cbdcbce01dea1ab3850be2311f33f00d75f6088b we introduced a Tim
Pope-inspired Git template scheme that automatically ran `ctags`
asynchronously through a hook after certain source control operations.
This hook is bypassed if a user uses `<leader>ct` to reindex their tags
from within Vim.

This change binds `<leader>ct` to use a Vimscript function that will
identify if the `ctags` hook exists in the local repository and executes
it. The function falls back to existing functionality if the hook does
not exist.

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -103,12 +103,6 @@ endfunction
 inoremap <Tab> <c-r>=InsertTabWrapper()<cr>
 inoremap <S-Tab> <c-n>
 
-" Exclude Javascript files in :Rtags via rails.vim due to warnings when parsing
-let g:Tlist_Ctags_Cmd="ctags --exclude='*.js'"
-
-" Index ctags from any project, including those outside Rails
-map <Leader>ct :!ctags -R .<CR>
-
 " Switch between the last two files
 nnoremap <leader><leader> <c-^>
 
--- a/vim/plugin/ctags.vim
+++ b/vim/plugin/ctags.vim
@@ -1,0 +1,15 @@
+" Exclude Javascript files in :Rtags via rails.vim due to warnings when parsing
+let g:Tlist_Ctags_Cmd="ctags --exclude='*.js'"
+
+" Index ctags from any project, including those outside Rails
+function! ReindexCtags()
+  let l:ctags_hook = '$(git rev-parse --show-toplevel)/.git/hooks/ctags'
+
+  if exists(l:ctags_hook)
+    exec '!'. l:ctags_hook
+  else
+    exec "!ctags -R ."
+  endif
+endfunction
+
+nmap <Leader>ct :call ReindexCtags()<CR>


vimrc
```diff
@@ -106,8 +106,18 @@ inoremap <S-Tab> <c-n>
...
+function! ReindexCtags()
+  let l:ctags_hook_location = '$(git rev-parse --show-toplevel)/.git/hooks/ctags'
+
+  if exists(l:ctags_hook_location)
+    exec '!'. l:ctags_hook_location
+  else
+    exec "!ctags -R ."
+  endif
+endfunction
+nmap <Leader>ct :call ReindexCtags()<CR>
```
gylaz: Thoughts on pulling this out into a file that is sourced here? Could be a nice step in the direction of breaking up `vimrc` into descriptively-named files.

geoffharcourt: I'm definitely interested in going in this direction. Are we going to place those files in `.vim/configs` (so this would go in `.vim/configs/ctags.vim`) and then source those files? I'm worried about placing them at the root level, but also want to cause the least possible disruption to anyone's existing configuration.

teoljungberg: I split my vimrc up by placing plugin configurations into `.vim/plugin/settings`, and extracting plugins like this one into `.vim/plugin`. That way, there is nothing that I need to source explicitly.

jkrmr: Just documenting here in case it's useful to anyone: When I switched to using the `plugin` folder instead of explicitly sourcing vimrc files, I ran into several mysterious instances where a plugin's settings refused to be overridden. I never quite nailed down why that was happening (presumably because I had less control over load order), but the `after` directory proved handy for working around these issues. 

I ended up storing all my general vim settings (leader key bindings, `ftdetect`, `ftplugin` settings, etc) in `~/.vim/after/plugin`, ensuring any intended overrides (e.g., of a plugin's default keybindings) will actually override plugins' default settings. I still use `~/.vim/plugin` for plugin config settings with files named after their plugins, as suggested by @christoomey.

```
:h after-directory

	The defaults for most systems are setup to search five locations:
	1. In your home directory, for your personal preferences.
	2. In a system-wide Vim directory, for preferences from the system
	   administrator.
	3. In $VIMRUNTIME, for files distributed with Vim.
							*after-directory*
	4. In the "after" directory in the system-wide Vim directory.  This is
	   for the system administrator to overrule or add to the distributed
	   defaults (rarely needed)
	5. In the "after" directory in your home directory.  This is for
	   personal preferences to overrule or add to the distributed defaults
	   or system-wide settings (rarely needed).

```

mike-burns: Wow, thank you for this, @jkrmr!

teoljungberg: @jkrmr Thanks for the detailed explanation! I agree, if you want to override plugins - `after` is the place to put it in.

PR #456, opened 2015-11-19
Run `db:test:prepare` task on `migrate` alias

geoffharcourt:
We use the `migrate` shell alias to run migrations, roll back one step,
then migrate again. In addition to serving as a handy shortcut for `rake
db:migrate`, it helps us test whether or not migrations are fully
reversible.

If a migration is rolled back and edited, then migrated forward, the
change may not be reflected in the test database, which can cause
unexpected errors when the development and test DB schema do not match.
This change runs the `db:test:prepare` task after migrations complete to
ensure that the test database schema matches the schema defined in the
application's `db/schema.rb` or `db/struture.sql`.

### DIFF
--- a/aliases
+++ b/aliases
@@ -9,7 +9,7 @@ alias v="$VISUAL"
 alias b="bundle"
 
 # Rails
-alias migrate="rake db:migrate db:rollback && rake db:migrate"
+alias migrate="rake db:migrate db:rollback && rake db:migrate db:test:prepare"
 alias s="rspec"
 
 # Pretty print the path


gylaz:
The commit description says `db:test:prepare`, yet the change adds `db:test:load`. What is the difference between the two commands?

geoffharcourt:
@gylaz corrected. I want to use `db:test:load` because it reloads the test schema from the application's defined schema. `db:test:prepare` checks for pending migrations in addition to loading the test schema:

http://stackoverflow.com/questions/7693365/whats-the-difference-between-dbtestclone-dbtestclone-structure-dbtestlo

croaky:
I feel like we used to rake db:test:prepare at the end this command. Does the Git history reveal why we removed it? Maybe Spring?

croaky:
Looks like Rails 4.1 deprecation was the reason: https://github.com/thoughtbot/dotfiles/commit/0b5eca112a5ea276ac85eef8392a51328f482580

croaky:
If test:prepare was not actually deprecated, it sounds like it has an advantage of working on .sql format schemas in addition to the Ruby format? Based on that Stack Overflow post.

I agree on trying to pull some kind of test database prep into this command.

geoffharcourt:
We removed `db:test:prepare` in https://github.com/thoughtbot/dotfiles/pull/260 because Rails 4.1 discouraged users from running the (at the time) deprecated Rake task. Rails brought back `db:test:load` in https://github.com/rails/rails/pull/17739 due to user complaints about being unable to force a test database synchronization.

Without dropping and recreating the test database from scratch, the current migration strategy of solely using `db:migrate` and `db:rollback` will never bring changes into test from a migration that was run, rolled back, modified, and then re-run. I often will do this as I'm working on a PR and I discover new requirements for a database column, in an attempt to keep the final merged migrations 1) as few in number as possible and 2) strictly organized around a particular functionality change for a model.

Running `db:test:load` or `db:test:prepare` on each `migrate` operation has a small performance penalty, but the task is only being run when you have a reason to want to check or force a synchronization of the database. I think knowing for sure that your test and development databases are at the same point in their evolution is worthwhile. 

Running `db:test:prepare` or `db:test:load` after running our existing `migrate` alias added just under 1.0 seconds on average to the `migrate` operation on an application with a reasonably sized DB schema.

geoffharcourt:
Given that we have some apps that will use `structure.sql` instead of `schema.rb`, it seems like `db:test:prepare` is the more flexible option as @croaky suggested.

croaky:
Looks good to me.

derekprior:
This is a great commit message (not that your original was bad either).

geoffharcourt:
@derekprior thanks for the recommendation, I've slightly edited my research comment and made that the commit message.

PR #457, opened 2015-11-24
[replace script] to be compatible with GNU/Linux

pedro-nonfree:
```sed -i ''``` is not working in Debian Jessie (I think GNU/Linux in general).
error:
```sed: can't read <sed expression>: No such file or directory```
sed (GNU sed) 4.2.2

What it works is:
```sed -i""``` or ```sed -i''``` or ```sed -i```

### DIFF
--- a/bin/replace
+++ b/bin/replace
@@ -9,4 +9,6 @@ shift
 replace_with="$1"
 shift
 
-ag -l --nocolor "$find_this" "$@" | xargs sed -i '' "s/$find_this/$replace_with/g"
+items=$(ag -l --nocolor "$find_this" "$@")
+temp="${TMPDIR:-/tmp}/replace_temp_file.$$"
+for item in $items; do sed "s/$find_this/$replace_with/g" $item > $temp && mv $temp $item; done

### CODE REVIEW

**Comment by mike-burns** [comment]
The only portable way to do this is with Perl.

**Comment by mike-burns** [comment]
Sorry for being terse there. Here's the most portable way to do this:

```
~% echo foo > bar              
~% perl -pi -e 's/foo/baz/' bar
~% cat bar                     
baz
```

Alternatively, drop the `-i` entirely and do the sed(1) & mv(1) dance.

**Comment by pedro-nonfree** [comment]
I totally agree with you to use perl

**Comment by geoffharcourt** [comment]
@mike-burns would we be better off distributing a `replace` script that doesn't use Perl, but uses mv(1) instead?

**Comment by mike-burns** [comment]
@geoffharcourt I'd prefer mv(1), in the least for those FreeBSD users out there. Your choice.

**Comment by geoffharcourt** [comment]
@pedro-nonfree want to take a crack at doing a more portable version using mv(1) and sed(1)? If we're going to make a change to improve portability, I'd rather go all the way and avoid Perl entirely.

**Comment by pedro-nonfree** [comment]
@geoffharcourt done

Is /tmp available in MacOS and FreeBSD? Seems yes

http://superuser.com/questions/581347/where-is-the-temp-folder-in-os-x
https://www.freebsd.org/doc/handbook/dirstructure.html

File: bin/replace [hunk 1]
```diff
@@ -9,4 +9,6 @@ shift
 replace_with="$1"
 shift
 
-ag -l --nocolor "$find_this" "$@" | xargs sed -i '' "s/$find_this/$replace_with/g"
+items=$(ag -l --nocolor "$find_this")
+temp="/tmp/replace_temp_file"
```
mike-burns: If you want to get fancy:

```sh
temp="${TMPDIR:-/tmp}/replace_temp_file.$$"
```

This way the user can override the `/tmp`, and adding the PID makes it less likely to conflict with an existing file.

(It seems that mktemp(1) is not in POSIX. Oh well.)

pedro-nonfree:
@mike-burns agree,
learning a lot with this PR :smile: 
anything else?

geoffharcourt:
@pedro-nonfree looks good to me. I'm going to do some testing on my own.

geoffharcourt:
@pedro-nonfree it appears that by removing the `"$@"` argument that was previously passed to `ag` that the script no longer makes use of the third argument to select the files to be processed.

geoffharcourt:
@pedro-nonfree this worked for me:

```shell
items=$(ag -l --nocolor "$find_this" "$@")
temp="${TMPDIR:-/tmp}/replace_temp_file.$$"
for item in $items; do
  sed "s/$find_this/$replace_with/g" $item > $temp && mv $temp $item;
done
```

pedro-nonfree:
I integrated all diffs in one commit, but this destroyed incremental commits of this PR. But, given the complexity, I think it's fine

geoffharcourt:
@pedro-nonfree I was going to squash everything down into one commit, so that's better.

geoffharcourt:
@pedro-nonfree: I made a couple small updates (surrounding variables in double-quotes, breaking the `for` loop up into multiple lines) and updated the commit message. Merged in as 2d9e8bf. Thank you very much for helping improve the script!

PR #459, opened 2015-12-08
Use one space, not two, after punctuation.

bernerdschaefer:
When using a join command (`gq`, `J`), vim defaults to adding two spaces
after '.', '?', and '!'.

Setting `nojoinspaces` causes vim to only insert a single space.

```
:help joinspaces

			*'joinspaces'* *'js'* *'nojoinspaces'* *'nojs'*
'joinspaces' 'js'	boolean	(default on)
			global
			{not in Vi}
	Insert two spaces after a '.', '?' and '!' with a join command.
	When 'cpoptions' includes the 'j' flag, only do this after a '.'.
	Otherwise only one space is inserted.
	NOTE: This option is set when 'compatible' is set.
```

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -68,6 +68,9 @@ set expandtab
 " Display extra whitespace
 set list listchars=tab:»·,trail:·,nbsp:·
 
+" Use one space, not two, after punctuation.
+set nojoinspaces
+
 " Use The Silver Searcher https://github.com/ggreer/the_silver_searcher
 if executable('ag')
   " Use Ag over Grep


croaky:
Interesting! I think of the "two space after punctuation" rule as a relic from the typewriter era. I guess we're using monospaced fonts in Vim, too, but I agree on the "one space" in the modern era.

jferris:
Nice.

geoffharcourt:
@bernerdschaefer this is a nice addition. Thanks!

PR #461, opened 2015-12-11
Remove existing PR message when creating a PR

geoffharcourt:
When creating a pull request with the `git-pr` command or `hub pull-request`, it is not possible to abort the creation of the PR by quitting Vim without writing in a similar fashion to aborting a commit with `:q!`.

Killing the Vim process succeeds in not creating the PR, but leaves behind the originally-built pull request content in `.git/PULLREQ_EDITMSG`. Upon switching branches or re-attempting to create the PR in the same branch, the content is already populated and is often incorrect.

This can be a rude surprise when attempting to create a single-commit PR and expecting the PR content to match your well-crafted commit message and `hub` suggests the contents of your last PR instead of your current commit.

This change deletes the left behind PR message before running `hub pull-request` to ensure that `hub` builds PR content from the current unmerged commits in the current branch.

### DIFF
--- a/bin/git-pr
+++ b/bin/git-pr
@@ -2,4 +2,4 @@
 
 set -e
 
-hub pull-request
+rm -rf "$(git rev-parse --git-dir)"/PULLREQ_EDITMSG && hub pull-request


DarthMike:
Shouldn't that be fixed or improved in hub itself?

geoffharcourt:
I looked into this on Hub, and the reason it exists is to retain the PR message if your initial PR creation attempt fails. Unfortunately it doesn't look possible to retain both functionalities without adding a command line argument to Hub or a second PR creation script with different behavior.

jferris:
You can use `:cq` to abort. This doesn't appear to leave around `.git/PULLREQ_EDITMSG`.

geoffharcourt:
Nice! Thanks, @jferris.

PR #462, opened 2015-12-15
Extract filetype-specific setup to ftplugin

geoffharcourt:
Vim's documentation advises that buffer-specific configuration based on
a filetype should be performed via a file in `$HOME/.vim/ftplugin`:
http://vimdoc.sourceforge.net/htmldoc/usr_41.html#41.12

This change moves Git commit message- and Markdown-related editor
settings into `ftplugin/gitcommit.vim` and `ftplugin/markdown.vim`, and
rewrites those settings to avoid using `autocmd`, letting Vim execute
`setlocal` when a file of the filetype is determined or set.

This organization of single-filetype-specific configurations could serve as
a template for language-specific settings such as https://github.com/thoughtbot/dotfiles/pull/449.

### DIFF
--- a/vim/ftplugin/css.vim
+++ b/vim/ftplugin/css.vim
@@ -1,0 +1,1 @@
+setlocal iskeyword+=-
--- a/vim/ftplugin/gitcommit.vim
+++ b/vim/ftplugin/gitcommit.vim
@@ -1,0 +1,3 @@
+" Automatically wrap at 72 characters and spell check commit messages
+setlocal textwidth=72
+setlocal spell
--- a/vim/ftplugin/markdown.vim
+++ b/vim/ftplugin/markdown.vim
@@ -1,0 +1,5 @@
+" Enable spellchecking
+setlocal spell
+
+" Automatically wrap at 80 characters
+setlocal textwidth=80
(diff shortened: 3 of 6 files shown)


JoelQ:
I like this change :+1: . It is much nicer than all those `autocmd` commands in the main `.vimrc`

teoljungberg:
I'm a fan of this :+1:

geoffharcourt:
@JoelQ @teo I was unable to determine if there's any startup time improvement.

Question: we have some remaining autocommands for local buffer settings that apply to multiple filetypes such as CSS/Less/Sass. Is there a way to move those into ftplugin without creating a separate `ftplugin/css.vim`, `ftplugin/sass.vim`, `ftplugin/less.vim`?

teoljungberg:
@geoffharcourt I've just let them sit in each file. But I guess we could create a `ftplugin/shared.vim`, or even put it into another folder so a filetype called `shared` won't load that, that is sourced from the others?

geoffharcourt:
@teoljungberg would sourced files be loaded and executed locally each time a buffer for a given filetype is opened?

I've pushed CSS/SCSS/Sass out to separate, one-line `ftplugin` files for now.

PR #463, opened 2015-12-19
Better match the default behavior of ctrl-P when using ag

sgrif:
When using ctrl-p without ag, the pattern given is not treated as a regular expression. Additionally, the value passed when first opening ctrl-p is not actually a valid PCRE regex, and produces an error when used on Windows. This modifies the options to treat the pattern as a string literal, not a regular expression, resolving the Windows problem, and better matching the behavior of ctrl-p without Ag.

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -64,7 +64,7 @@ if executable('ag')
   set grepprg=ag\ --nogroup\ --nocolor
 
   " Use ag in CtrlP for listing files. Lightning fast and respects .gitignore
-  let g:ctrlp_user_command = 'ag %s -l --nocolor --hidden -g ""'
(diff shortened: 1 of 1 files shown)


geoffharcourt:
@sgrif, this improves the consistency of file search in Ctrl-P with and without Ag, but if I'm understanding this correctly does it remove an existing capability (to use a PCRE regex as your search term) for Ag users?

sgrif:
Correct. IMO the behavior of a vim plugin shouldn't change based on an
executable being present on a machine.

jferris:
This seems good to me. I didn't even realize I could use regular expressions with ctrl-p, and I've probably only done so by mistake. We can always deliberately enable regular expressions with local configuration if we want to.

PR #464, opened 2016-01-11
Globally ignore compiled Python files.

georgebrock:
The first time a `.py` file is run, or any time it's been modified, Python creates a `.pyc` file containing a bytecode representation.

These bytecode files should always be ignored by Git, since keeping them updated in a repo would be time consuming and would serve no purpose.

### DIFF
--- a/gitignore
+++ b/gitignore
@@ -9,3 +9,4 @@ tags
 !tags/
 tmp/**/*
 !tmp/cache/.keep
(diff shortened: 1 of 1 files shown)


geoffharcourt:
LGTM

georgebrock:
Thanks, Geoff. Merging.

PR #467, opened 2016-02-07
Add `install-missing-ruby` alias

croaky:
Sometimes I don't have installed on my machine
the Ruby version that the project requires.
Almost every project uses the `.ruby-version` convention.
I type `install-missing-ruby` (or tab complete a few characters)
and the version of Ruby is installed.

### DIFF
--- a/aliases
+++ b/aliases
@@ -5,10 +5,9 @@ alias mkdir="mkdir -p"
 alias e="$EDITOR"
 alias v="$VISUAL"
 
-# Bundler
+# Ruby
 alias b="bundle"
-
-# Rails
+alias install-missing-ruby="brew update && brew upgrade ruby-build && rbenv install"
 alias migrate="rake db:migrate db:rollback && rake db:migrate db:test:prepare"
 alias s="rspec"
 


geoffharcourt:
I really like this idea.

From what I can tell we've avoided configuration in dotfiles that depends on a certain platform. This alias assumes that the user is on OS X, has Homebrew installed, and uses rbenv/ruby-build. Would a zsh function that returns a message if these conditions (or at least the first two) aren't true be a better fit?

benoittgt:
It seems to me too dependent to platform you use. I think it deserves a separate script but not in this repo.:smile:  

What about people who want to use rvm ?

geoffharcourt:
@benoittgt as a team we've standardized around rbenv in dotfiles, but there's no reason you couldn't write your own rvm-compatible replacement and put that in your `.aliases.local`.

This change would mark the first time we've done something OS X-specific in dotfiles (or at least the first reintroduction of such a setting from a currently agnostic state). I think it's worth discussing whether we want to do that even if the addition is not a net-negative for non-OS X users.

JoelQ:
I like the idea here but I don't know that it fits with the philosophy of the dotfiles since it is very coupled to a very specific Ruby setup on a particular operating system.

mike-burns:
This would make more sense under an `osx` tag.

geoffharcourt:
@croaky do you want to put this under the `osx` tag as @mike-burns suggested above? I think at that point we'd be OK to merge.

croaky:
Happy to move this to an OS X tag. I unfortunately don't understand what that means, though. Are tags a dotfiles or RCM thing? Is there some documentation that could be added to dotfiles' CONTRIBUTING.md explaining tags? Possibly a good GR article, too.

geoffharcourt:
@croaky, tags are RCM's way of [selectively including certain configuration files on certain machines](http://thoughtbot.github.io/rcm/rcup.1.html).

@gfontenot has a [nice set of tags in his dotfiles](https://github.com/gfontenot/dotfiles).

If you put this change into a directory named `tag-osx`, it will only be symlinked by `rcup(1)` if that machine's `rcrc` (or options passed to `rcup(1)` specify the `osx` tag.

calebthompson:
The "official" recommendation may be to use one ruby manager, but plenty of us use others. I've always had better luck with rvm.

geoffharcourt:
I've started using this alias on my own machine, but I think it's too bound to the choices of OS X and rbenv to use in dotfiles. After further reflection I don't think tags help us, as tags aren't a solution for multiple options that would all need to share a single file. (I guess we could optionally source an `aliases.osx` file, but I'm not sure if we're ready for that.)

PR #469, opened 2016-02-11
Remove aliases that are no longer included

acconrad:
The README points to some aliases that are no longer included in the `.aliases` file, and should be removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,8 +196,6 @@ configuration:
 Shell aliases and scripts:
 
 * `b` for `bundle`.
-* `g` with no arguments is `git status` and with arguments acts like `git`.
-* `git-churn` to show churn for the files changed in the branch.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -196,8 +196,6 @@ configuration:
 Shell aliases and scripts:
 
 * `b` for `bundle`.
-* `g` with no arguments is `git status` and with arguments acts like `git`.
```
gylaz: We still have this as a function: https://github.com/thoughtbot/dotfiles/blob/master/zsh/functions/g

README.md
```diff
@@ -196,8 +196,6 @@ configuration:
...
-* `git-churn` to show churn for the files changed in the branch.
```
gylaz: This also exists: https://github.com/thoughtbot/dotfiles/blob/master/bin/git-churn

PR #470, opened 2016-02-12
Use vim-test instead of vim-rspec

derekprior:
vim-rspec is an excellent Vim plugin for running RSpec tests but there
are many of us who run different kinds of tests on any given day.
vim-test is built to support a wide array of test frameworks and
strategies for executing those tests.

vim-test ships with support for several languages and frameworks in use
on thoughtbot projects today:

* Ruby (both RSpec and Minitest)
* Elixir (ExUnit and espec)
* Go (gotest)
* Python (nose, pytest, djangotest, djangonose)
* JavaScript (jasmine, mocha)

It also supports several execution strategies I have seen people use at
thoughtbot:

* Dispatch
* vim-tmux-runner
* Tslime
* Terminal.app/iTerm.app tabs
* neovim's native terminal
* `:!`

vim-test is extensible, both for additional languages/frameworks and for
execution contexts. See: https://github.com/janko-m/vim-test


benoittgt:
Thanks @derekprior . Didn't know *vim-test*. Looks good to me. I'm using only rspec and test-unit but it works perfectly on both.

jferris:
This seems like a good idea. Is anybody using this internally right now?

derekprior:
I've been using it for a couple weeks now. It's worked well for both elixir and ruby.

derekprior:
It does not natively support whatever ember-cli tests are though. That might be a good test of extensibility, but I barely know how to run those tests manually, so I'm not the best candidate to tackle it.

jferris:
> It does not natively support whatever ember-cli tests are though.

Neither does vim-rspec, so I don't think that should be a problem with getting this merged.

geoffharcourt:
@derekprior this worked great in my testing with Neovim (ran tests by default in a spawned terminal even without tmux), vim with vim-dispatch, and vim without other plugins.

In the absence of any reported issues, I'm going to merge.

geoffharcourt:
Merged as 87de45b0c61b38ea288f992ad321fdb7b650d144.

PR #472, opened 2016-02-18
Add rhubarb.vim for fugitive Github support

geoffharcourt:
We do most of our client work on Github,
and a large portion of our personal projects
are hosted there.

fugitive.vim currently supports
opening the current file in the current branch
in a browser at Github through the `Gbrowse` command.
That command is [slated to be removed to make fugitive
agnostic about coding sites][fugitive-docs].

The capability to quickly generate or view a Github URL
specific to the current branch
can be useful in PR reviews
or other discussions about code.

This change adds rhubarb.vim,
a plugin that implements the slated-to-be-removed
functionality of fugitive that works with Github.
It also provides the user with omni-completion (`<C-x><C-o>`)
to autocomplete Github issues, issue URLs, and collaborators.

[fugitive-docs]: https://github.com/tpope/vim-fugitive/blob/18d6d1ab82d9ac15586d7d3c1a36f9ef6fb50eae/doc/fugitive.txt


MattMSumner:
This seems like a great addition to me. :+1:

seanpdoyle:
[We already have `hub browse`](https://github.com/thoughtbot/laptop/blob/66f16d8690161fbe72b884c4044f4354d83b7fac/mac#L163-L164) available to us.

I've also aliased `git` to use `hub` (others might've as well).

Is adding support for `:Gbrowse` through a plugin worthwhile, when we can already `:!hub browse` or `:!git browse` or <kbd>ctrl</kbd> + <kbd>z</kbd> `git browse`?

geoffharcourt:
I think rhubarb's `:Gbrowse` command is an improvement over `hub browse` because it leverages what you're looking at in Vim. Selecting a range in visual mode then using `Gbrowse` as `:'<,'>Gbrowse` opens the file in the current branch on Github, with the link jumping directly to the line numbers in question, highlighting those lines on Github. It makes it much easier to point your collaborators to the code in question than to have to separately specify the line number.

From the hub docs and a couple minutes of testing, it doesn't look like `hub browse` opens a specific file, just the repo in the current branch (`hub browse` can jump right to the PR list or issues list or even the wiki which I hadn't realized)

If this is too heavy-handed I'm fine with leaving it out, but I think being able to quickly generate links that point to a specific piece of code would be a worthwhile convenience.

calebthompson:
I agree with Geoff, but Sean you can also `:Git <any subcommand> <whatever arguments>` so no need to shell out.

geoffharcourt:
Thanks everyone for the feedback.

PR #473, opened 2016-02-22
Update copyright date

geoffharcourt:
It's 2016.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,7 +2,7 @@ LICENSE
 
 The MIT License
 
-Copyright (c) 2009-2014 thoughtbot, inc.
+Copyright (c) 2009-2016 thoughtbot, inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


blainesch:
The [LICENSE](https://github.com/thoughtbot/dotfiles/blob/master/LICENSE#L5) file has a date too.

geoffharcourt:
Which was even more out of date! Thanks for the spot @blainesch!

bernerdschaefer:
IANAL, but based on http://www.copyright.gov/circs/circ03.pdf and http://www.copyright.gov/circs/circ01.pdf, it seems that only the original date of publication is needed (if even that).

What about making it "Copyright (c) 2009 thoughtbot", so the dates can't fall behind?

geoffharcourt:
@bernerdschaefer I was hoping there was a good cause to remove it entirely, but this works too. My only concern is that having a date but having it be quite a while in the past may make the project look unmaintained at a quick glance. Am I being paranoid?

bernerdschaefer:
@geoffharcourt What about merging this as-is, and then starting a discussion on thoughtbot/guides so we can tackle this for multiple projects?

geoffharcourt:
:+1: Thanks @bernerdschaefer and @blainesch for the review and :eyes:

PR #474, opened 2016-02-26
Add hub support to the g function

jeffcole:
Since the [`g` function](https://github.com/thoughtbot/dotfiles/blob/master/zsh/functions/g) doesn't have access to a [`git='hub'` alias](https://github.com/github/hub#aliasing), we can't run `hub` commands through it.

This checks for `hub` in the `g` function and uses it instead of `git` in the event that it's found.

It also adds a similar check for `hub` before invoking it in [`git-pr`](https://github.com/thoughtbot/dotfiles/blob/master/bin/git-pr).

### DIFF
--- a/bin/git-pr
+++ b/bin/git-pr
@@ -2,4 +2,12 @@
 
 set -e
 
-hub pull-request
+if command -v hub >/dev/null; then
+  hub pull-request
+else
+  cat <<MSG >&2
+This command depends upon 'hub', which was not found.
+Visit https://hub.github.com for installation instructions.
+
+MSG
+fi
--- a/zsh/functions/g
+++ b/zsh/functions/g
@@ -1,9 +1,18 @@
 # No arguments: `git status`
 # With arguments: acts like `git`
+# Uses `hub` if available
 g() {
-  if [[ $# -gt 0 ]]; then
-    git "$@"
+  local _git
+
+  if command -v hub >/dev/null; then
+    _git=hub
   else
-    git status
+    _git=git
+  fi
+
+  if [[ $# -gt 0 ]]; then
+    $_git "$@"
+  else
+    $_git status
   fi
 }


geoffharcourt:
I'm [a fan of doing this](https://github.com/geoffharcourt/dotfiles-local/blob/master/zsh/functions/g). One question I have is whether checking for the command incurs any non-trivial waiting. Since the `g` function gets run so often, anything that causes it to have a delay would be unwelcome.

cc: @thoughtbot/shell any insight here?

mike-burns:
I don't have any insight here except that you should be using gitsh. Maybe someone else from @thoughtbot/shell has more thoughts on these functions in particular.

jeffcole:
@geoffharcourt I like the approach of overriding the function in local dotfiles, except that it represents an (admittedly trivial) divergence from the org's dotfiles — that might be avoided since most users probably have `hub`.

It seems that these dotfiles should either be explicit about requiring `hub`, gracefully handle the case where it doesn't exist, or remove usage. The goal with this PR is to provide for the second case while following the spirit of `hub`'s pass-through usage.

Unfortunately I don't have insight into how much delay is added, but agree that any non-trivial amount would be bad news. I'm mostly just curious what y'all think about the above :point_up:

Thanks for the project and the feedback!

jeffcole:
I forgot to mention that I have been using this change for a bit under a week and haven't noticed any perceptible delay.

Thanks again!

jferris:
I personally don't use `hub` as an alias for Git. I use `git pr` as an alias for `!hub pull-request`, which makes it possible to do `g pr` or just `pr` from a `gitsh` session.

I would prefer to keep a simpler `g` alias and use `gitsh` when I need to do any meaningful work with Git.

geoffharcourt:
I think we have enough split on using `git`, `hub`, and `gitsh` here that this makes sense to maintain as a personal customization rather than a team-wide feature. 

@jeffcole thanks for the idea.

jeffcole:
@geoffharcourt @jferris @mike-burns thanks for the feedback!

PR #475, opened 2016-03-22
Fix typographical error

gfvcastro:


geoffharcourt:
@gfvcastro our intention is to leave the British English spelling here as a salute to our colleagues in [thoughbot's London office](https://thoughtbot.com/london).

PR #478, opened 2016-03-31
Avoid expanding tabs to spaces for Golang files

gylaz:
* `noexpandtab` will ensure tabs are used instead of staces when tabbing
or auto-indenting.
* `softtabstop=2` isn't needed because we are already setting
`tabstop=2` in our `.vimrc`.

### DIFF
--- a/vim/ftplugin/go.vim
+++ b/vim/ftplugin/go.vim
@@ -1,6 +1,6 @@
 let g:go_fmt_command = "goimports"
 
-setlocal softtabstop=2
 setlocal listchars=tab:\ \ ,trail:·,nbsp:·
+setlocal noexpandtab
 
 compiler go


gylaz:
ping @thoughtbot/go

gylaz:
ping @thoughtbot/go

bernerdschaefer:
I don't know enough to weight in on the `shifttabstop` change, but it looks like I have `noexpandtab` set in my own Go configuration so :+1: there.

gylaz:
From [this](http://vim.wikia.com/wiki/Indenting_source_code) it looks like `softtabstop` is necessary if spaces are desired, otherwise `tabstop` should be set with `noexptandtab` and `softtabstop` left unset.

geoffharcourt:
Thanks @gylaz!

PR #479, opened 2016-04-07
Overwrite internal field separator when replacing.

blainesch:
Problem:
Currently the `for` in bash will break up lists by whitespace, so if a
filename has a space in it the `for` will break it up incorrectly.

AG will separate the files with a newline character, so this temporarily
overwrites the `IFS` to be used correctly for this use case.

[Docs](https://bash.cyberciti.biz/guide/$IFS)

### DIFF
--- a/bin/replace
+++ b/bin/replace
@@ -11,6 +11,7 @@ shift
 
 items=$(ag -l --nocolor "$find_this" "$@")
 temp="${TMPDIR:-/tmp}/replace_temp_file.$$"
+IFS=$'\n'
 for item in $items; do
   sed "s/$find_this/$replace_with/g" "$item" > "$temp" && mv "$temp" "$item"
 done


gylaz:
Do we want to unset `IFS` after the script did its job, or is the variable only set for the lifetime of the script/command?

geoffharcourt:
cc: @thoughtbot/shell

blainesch:
In my tests this is only overwritten in the script. However, I've only tested this on El Captan.

~~~
▶ cat test.sh 
IFS=$'\n'

echo $IFS | md5

▶ echo $IFS | md5
5d96df8e49252b906e7d64df165f32d9

▶ ./test.sh 
68b329da9893e34099c7d8ad5cb9c940

▶ echo $IFS | md5
5d96df8e49252b906e7d64df165f32d9
~~~

mike-burns:
This should be fine.

For clarity, maybe move the assignment above the `for` loop.

blainesch:
@mike-burns updated.

geoffharcourt:
Merged as 1dc550d9850337594401349ea0ceeb922beef405.

Thanks for the contribution, @blainesch!

blainesch:
Thanks!

Issue #480, opened 2016-04-07
Maintaining a fork


blainesch:
So the problem exists when I try and pull in changes. The workflow I use at work for updating a branch would usually involve a rebase to avoid merge commits. This is problematic in this case, since a rebase would remove my gitconfig then the rebase would fail due to not having a gitconfig.

The best solution I have found is to squash my commits, which means I could potentially lose history I might end up reverting later. After squashing I simply `fetch thoughtbot` then `reset --hard thoughtbot/master` then `cherry-pick` my single commit on top. This seems like odd behavior that this works, since it still recreates the commit (asserting based on a new sha1) but doesn't complain like the rebase does.

The same problem exists when I want to create a pull request, since you obviously don't want my gitconfig, I usually end up re-cloning the repo to make the changes. I've not played around much with [git worktree](https://git-scm.com/docs/git-worktree) but that could be a solution.

* There must be a better way of handling this.
* It would be great if this repo provided a solution, either as a `bin` script, custom git command, or at least docs on one solution.

Having a recommended solution could improve contributions, and approachability/adoption of the repo.

geoffharcourt:
@blainesch are you making local changes that are just for you? Are those things that could be handled using `.local` files and/or your own local dotfiles repo with `rcm`?

For example, in the gitconfig situation you referred to I'm making overrides in my personal `gitconfig.local` rather than editing the team-shared thoughtbot `gitconfig` here: https://github.com/geoffharcourt/dotfiles-local/blob/master/gitconfig.local

blainesch:
@geoffharcourt that's correct, but those are committed to the repo so it can be shared across multiple computers (work/home). Is this not the expected use case?

blainesch:
@geoffharcourt oh! I see, you are keeping those in a [separate repo](https://github.com/geoffharcourt/dotfiles-local)! I didn't realize the `dotfiles` supported that. After digging around it's been in place for [a while](https://github.com/thoughtbot/dotfiles/commit/9a2cb782b151428711921dd6693da4365ce8c9c1) now. Thanks for pointing it out!

blainesch:
Actually [this](https://github.com/thoughtbot/dotfiles/commit/62a2b5eb8ff98fed9b9ca6a7254e16beef3a3f05) is a better representation of when it was implemented.

gylaz:
[...]

PR #481, opened 2016-04-07
Make it clear that customizations can be placed in their own directory.

blainesch:
I was placing my `.local` files inside my fork of this dotfiles which
made it harder to work with. This update to the instructions lets users
know they can version their local files separately.

Related to #480

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ any new files in the repository.
 Make your own customizations
 ----------------------------
 
-Put your customizations in dotfiles appended with `.local`:
+Put your customizations in `~/dotfiles-local` appended with `.local`:
 
 * `~/.aliases.local`
 * `~/.git_template.local/*`


geoffharcourt:
We're already stating that the default location for customizations with `rcm` in [the Install notes](https://github.com/thoughtbot/dotfiles/blob/248fb063325d5650c0b47aef05b01a29a054ffb7/README.md#install). 

I do understand that there's a lot of room for better clarity here. Can we revise this to emphasize that local changes shouldn't be placed in the repo, but should either be directly in `$HOME` or in a separate directory?

README.md
```diff
@@ -53,7 +53,7 @@ any new files in the repository.
 Make your own customizations
 ----------------------------
 
-Put your customizations in dotfiles appended with `.local`:
+Put your customizations in `~/dotfiles-local` appended with `.local`:
```
gylaz: I think this is too prescriptive in regards to path and directory. How about just stating that customizations can be up in **a directory** and added to `rcrc`?

blainesch: Since there are a few methods to maintaining customizations, what about having a section "Maintaining your customizations" that explains forking the dotfiles, a dotfiles-local and just managing the changes locally without a vcs?

PR #482, opened 2016-04-14
Add a command to fixup your commit.

blainesch:
I was originally thinking `git fixup` would be good here, but I see there's two letter standard going on.

### DIFF
--- a/gitconfig
+++ b/gitconfig
@@ -9,6 +9,7 @@
   ap = add --patch
   branches = for-each-ref --sort=-committerdate --format=\"%(color:blue)%(authordate:relative)\t%(color:red)%(authorname)\t%(color:white)%(color:bold)%(refname:short)\" refs/remotes
   ci = commit -v
+  fixup = commit --fixup
   co = checkout
   pf = push --force-with-lease
   st = status


mike-burns:
Is this something we do often? Is this something I should be doing more often?

calleerlandsson:
I use it whenever I add a commit to a branch with the intention of squashing it into another before merging the branch. `commit --fixup SHA` is made to work with with `rebase --interactive --autosquash`.

geoffharcourt:
I started doing this after reading @georgebrock's [excellent blog post](https://robots.thoughtbot.com/autosquashing-git-commits) on autosquashing. It's particularly nice when you want to temporarily submit a commit (that will later be squashed or fixed up) to suggest a resolution to a PR reviewer's comments without rebasing and changing commit references while a review might still be in progress.

jsteiner:
If everyone likes this, it might be nice to have `sq = commit --squash` in there as well.

mike-burns:
Is this something we do so often that the short `fu` and `sq` aliases make more sense than `fixup` and `squash` ?

blainesch:
The [autosquash flag which is on](https://github.com/thoughtbot/dotfiles/blob/master/gitconfig#L25) already inside of the global `gitconfig`.

blainesch:
I have updated to include both `squash` and `fixup` as commands.

gylaz:
I'd like to push back on this. In a normal workflow process we don't type this enough to save us a few characters. I think our dotfiles should be slim enough to be accessible to new-comers, rather than bulk up with all of the possible commands we *might* use during the day. My $0.02.

blainesch:
Amend actually already exists as `git ca` I realized as I was typing it.

There was a level of indirection, since it lives in bin though.

https://github.com/thoughtbot/dotfiles/blob/master/bin/git-ca

blainesch:
I've updated the commit to just have `fixup`.

geoffharcourt:
Hi @blainesch I think there's enough disagreement on this that I'm going to close without merging. (However I'll probably be pulling this into my own overrides.) Thanks for your contribution(s) to dotfiles.

PR #483, opened 2016-04-15
Git ignore .byebug_history

tute:
After using byebug on any Ruby file, the sessions history is written to
this file. It shouldn't be part of our repositories, so we ignore it.

See also: https://github.com/thoughtbot/suspenders/pull/710, where @ivobenedito started work on this.


gylaz:
I have this in my own dotfiles. 👍

PR #484, opened 2016-04-25
Add a zsh binding for push-line-or-edit

mike-burns:
The use case for push-line-or-edit:

1. You are in the middle of typing a long command, perhaps something
involving `tar`.
2. You realize that you must first run another command, such as `man
tar`.
3. You invoke `push-line-or-edit`, which gives you a fresh prompt. You
type `man tar` and read as needed.
4. When the command (`man tar`) finishes, the long command you
half-typed is back, exactly as you left off.

Other use cases are realizing that you're in the wrong directory before
you press enter, changing your Ruby version before you press enter on
that `bundle` command, or in general being partway through something and
realizing that you're not quite ready to run it yet.

I picked <kbd><kbd>CTRL</kbd><kbd>B</kbd></kbd> arbitrarily; let me know if you have a more natural keybinding.

### DIFF
--- a/zsh/configs/keybindings.zsh
+++ b/zsh/configs/keybindings.zsh
@@ -1,3 +1,6 @@
+# give us access to ^Q
+stty -ixon
+
 # vi mode
 bindkey -v
 bindkey "^F" vi-cmd-mode
@@ -10,4 +13,5 @@ bindkey "^R" history-incremental-search-backward
 bindkey "^P" history-search-backward
 bindkey "^Y" accept-and-hold
 bindkey "^N" insert-last-word
+bindkey "^Q" push-line-or-edit
 bindkey -s "^T" "^[Isudo ^[A" # "t" for "toughguy"


zsh/configs/keybindings.zsh
```diff
@@ -10,4 +10,5 @@ bindkey "^R" history-incremental-search-backward
 bindkey "^P" history-search-backward
 bindkey "^Y" accept-and-hold
 bindkey "^N" insert-last-word
+bindkey "^B" push-line-or-edit
```
teoljungberg: `^B` collides with an already existing mapping, what do you think of using another mapping? Like `^Q` that was mentioned on Slack.

jferris: `^B` also collides with the default tmux leader key. I found a [reference card](http://www.bash2zsh.com/zsh_refcard/refcard.pdf) (warning: PDF) that lists the default zle bindings in each mode (page13), which may be useful.

mike-burns: I see that we're using <kbd><kbd>^</kbd><kbd>s</kbd></kbd> for our secondary tmux prefix. Do we disable XON/XOFF flow control somewhere in dotfiles that I'm not seeing? If not, how is that prefix working?

<kbd><kbd>^</kbd><kbd>q</kbd></kbd> would also require XON/XOFF.

The default zle bindings have mapped this key to <kbd><kbd>ESC</kbd><kbd>q</kbd></kbd> for emacs mode and nothing for vi mode. We use vi mode. The emacs mode mapping, due to its use of <kbd>ESC</kbd>, would not work in vi mode.

mike-burns:
No consensus, but if you pick one I'll gladly change it to that.

geoffharcourt:
My first five attempts at something with a mmemonic match failed.

<kbd>Ctrl+j</kbd> looks like it's free? There aren't many zle bindings that are unclaimed. We could do <kbd>Ctrl+q</kbd> and let users decide if they'd rather have flow control on or use this feature.

mike-burns:
<kbd><kbd>ctrl</kbd><kbd>j</kbd></kbd> is the same as <kbd>Enter</kbd>. I'll change this to <kbd><kbd>ctrl</kbd><kbd>q</kbd></kbd> and turn off flow control in dotfiles.

PR #485, opened 2016-05-03
Use vim-polyglot instead of individual syntaxes

sheerun:
Hey, I thought I'd propose to use my syntax plugin instead of multiple syntaxes. You can read the advantages here: https://github.com/sheerun/vim-polyglot

I maintain plugin for more than 3 years and it doesn't seem to change in near future.

### DIFF
--- a/vimrc.bundles
+++ b/vimrc.bundles
@@ -29,12 +29,10 @@ call plug#begin('~/.vim/bundle')
 " Define bundles via Github repos
 Plug 'christoomey/vim-run-interactive'
 Plug 'ctrlpvim/ctrlp.vim'
-Plug 'fatih/vim-go'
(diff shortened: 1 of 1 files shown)


blainesch:
This removes 3 packages in the core, and adds 70 indirectly. Looking at the list of syntax highlighting I've installed in my `dotfiles-local` this isn't a full solution.

geoffharcourt:
Hi @sheerun, thanks for your PR.

Given recent discussions we've had about the size of dotfiles and what it installs on machines, I don't think vim-polyglot is a good fit. As a group our current approach is to consider individual packages carefully before putting them into everyone's coding environment. There's a lot going on in polyglot. The scope of what's going on there is probably too large for us to include.

sheerun:
👍

PR #486, opened 2016-05-23
Configure ctags for javascript projects

iwz:


ctags
```diff
@@ -13,3 +13,11 @@
...
 --regex-Elixir=/^[ \t]*Record\.defrecord[ \t]+:([a-zA-Z0-9_]+)/\1/r,records,records (defrecord...)/
 --regex-Elixir=/^[ \t]*test[ \t]+\"([a-z_][a-zA-Z0-9_?! ]*)\"*/\1/t,tests,tests (test ...)/
+--exclude=bower_components
+--exclude=node_modules
+--exclude=vendor
+--languages=-javascript
+--langdef=js
+--langmap=js:.js
+--langmap=js:+.jsx
+--regex-js=/[ \t.]([A-Z][A-Z0-9._$]+)[ \t]*[=:][ \t]*([0-9"'\[\{]|null)/\1/n,constant/
```
teoljungberg: Are there more things we need to configure? What else doesn't work out of the box?

iwz: There are, apparently, more things we need to configure. @BlakeWilliams shared http://dance.computer.dance/posts/2015/04/using-ctags-on-modern-javascript.html which outlines a bunch more things that don't work out of the box. For example, ES2015 classes and their methods are not properly found, even with this PR. There's more work to do here

teoljungberg: Could the commit message reflect that?

gylaz:
Is https://github.com/thoughtbot/suspenders/blob/master/templates/dotfiles/.ctags a better fit for this? I realize that Suspenders is Rails specific, but are we doing enough Node, to warrant everyone who uses dotfiles to have these tags?

iwz:
@gylaz I don't know if we're doing enough Node at thoughtbot to warrant this addition. But there is Elixir-specific config in this file already. And for the past six months I've been working with Node and React on the projects I've been on. Perhaps that's enough?

gylaz:
I'm not a fan of having Elixir tags in here either (I haven't done an Elixir project, but I carry around this baggage). I'd rather us look for a better approach than use a precedent. I may be overreacting, but I fall into the camp of keeping our dotfiles a slim as possible.

I think a good question to answer here is, does this solve a pain-point for enough people to warrant being add to thoughtbot's `dotfiles`?

jferris:
Is it possible for us to split out directories or separate repositories for supporting different languages? I know RCM supports some of this via tags.

If we're trying to keep the dotfiles slim, doesn't JavaScript support still make sense? It's one of our core competencies, alongside Ruby.

gylaz:
>  doesn't JavaScript support still make sense? It's one of our core competencies, alongside Ruby.

My question is, does it belong here, in `dotfiles`? Or does it belong in `suspenders`, or the Node equivalent (should we have one?)?

derekprior:
As far as I know, this baggage costs nothing. Never use Elixir or JavaScript and this doesn't hurt. It doesn't take up keybindings, impact editor load times, or anything like that.

RCM tags are an interesting idea that's come up a couple times but no one has bit yet. I wish they worked in more cases, however. For instance, your ctag configuration has to be a single file. You can't load another file from within it. How would I opt in to Elixir and JavaScript ctag settings?

mike-burns:
Relevant rcm issue: https://github.com/thoughtbot/rcm/issues/166

iwz:
Merging this. If we have a good solution that allows us to remove Elixir and Javascript configs from, we can remove them both at the same time.

andyw8:
@iwz Just curious, how does the `--exclude=vendor` entry relate to JS? I'm working on a Rails project which has some 'internal' gems and engines in the `vendor` directory, and I would prefer ctags still to index them.

PR #490, opened 2016-06-23
Keep swap files in case of crash

dijonkitchen:
Reference: http://vi.stackexchange.com/questions/177/what-is-the-purpose-of-swap-files

Instead, just set *.swp to be ignored in .gitignore_global

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -4,7 +4,6 @@ let mapleader = " "
 set backspace=2   " Backspace deletes like most programs in insert mode
 set nobackup
 set nowritebackup
-set noswapfile    " http://robots.thoughtbot.com/post/18739402579/global-gitignore#comment-458413287
 set history=50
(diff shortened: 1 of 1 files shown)


mikegee:
Hrm. That link used to point to a particular comment on the blog post justifying `noswapfile`, but I don't see any comments on that post.

dijonkitchen:
Yeah, I realized that too. But I couldn't find evidence to the contrary.

geoffharcourt:
Hi @dijonkitchen, thanks for your contribution to dotfiles!

There's clearly some benefit to having and not having swap files. At the moment I think a number of our team like not having swap files for reasons of removing clutter and performance in Vim when loading large files. Since it's fairly easy for users to override this particular setting in their `vimrc.local`, this seems like a setting that users can opt-in to if they want.

PR #491, opened 2016-06-28
Update install documentation

### LINKED ISSUE #488: explicitly state you must clone dotfiles to ~
I tried to follow the directions [here](https://github.com/thoughtbot/dotfiles#install) and could not get it installed because I had my clone of dotfiles not located in the home directory. If there is no other way to get this to work correctly, you should explicitly state this, changing this line:

"Clone onto your laptop:"
to
"Clone onto your laptop in your home directory:"

This is very confusing to new users that misunderstand this because when you run `rcup` there is no helpful message telling you what's going on, it just returns nothing.

geoffharcourt:
In #488 there was some confusion over where to clone the dotfiles
repository before running `rcm`. While the repository can be placed
anywhere, by default users should be placing it in `~/dotfiles` if they
are using the default installation instructions and an unmodified
`.rcrc`.

Close #488.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Install
 
 Clone onto your laptop:
 
-    git clone git://github.com/thoughtbot/dotfiles.git
+    git clone git://github.com/thoughtbot/dotfiles.git ~/dotfiles
 
 (Or, [fork and keep your fork
 updated](http://robots.thoughtbot.com/keeping-a-github-fork-updated)).


jdbann:
It took me a while to make sense of setting up my .local files. Would it be worth adding in some explanation about handling .local files in a repository alongside the latest thoughtbot/dotfiles version?

I'd be happy to document for the wiki?

mike-burns:
@jdbann that'd be a great PR against the README, I think.

We have a blog post on the topic that might be useful to link to: https://robots.thoughtbot.com/manage-team-and-personal-dotfiles-together-with-rcm

PR #493, opened 2016-07-05
Use "absolute" paths for agignore

tsigo:
This seems to resolve an issue where CtrlP would randomly return "NO
ENTRIES".

See https://forum.upcase.com/t/ctrlp-sometimes-displays-no-entries/5658

This also sorts the paths alphabetically, because OCD. :tada:

### DIFF
--- a/agignore
+++ b/agignore
@@ -1,7 +1,7 @@
-log
-tags
-tmp
-vendor
 .git/
-bower_components/
-node_modules/
+tags
+/bower_components/
+/log
+/node_modules/
+/tmp
+/vendor


geoffharcourt:
I don't use CtrlP, but I tested this and it seemed to work as expected.

Is this a bug with Ag? Would it be worth creating a failing scenario that doesn't involve Vim and reporting it there?

tsigo:
> Would it be worth creating a failing scenario that doesn't involve Vim and reporting it there?

It would be, but I can't be certain it's a bug with Ag itself, as I've only seen this happen in the context of CtrlP [calling out to `ag`](https://github.com/thoughtbot/dotfiles/blob/85b11b658322e2ed4ea747aabea126977b55c789/vimrc#L71).

Anecdotally, I used to see that "NO ENTRIES" error dozens of times a day and haven't seen it at all since this change.

geoffharcourt:
From discussion with other developers, it seems this problem happens with repositories that have a very large number of files.

Merged as c0717e7. Thank you for your contribution!

PR #497, opened 2016-07-22
Add Ag mappings in vim

geoffharcourt:
https://robots.thoughtbot.com/faster-grepping-in-vim

- Create an `:Ag` command if none is defined to perform a search with `Ag` and render results in quickfix window
- map `\` in normal mode to set up the command for an argument

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -72,6 +72,11 @@ if executable('ag')
 
   " ag is fast enough that CtrlP doesn't need to cache
   let g:ctrlp_use_caching = 0
+
+  if !exists(":Ag")
+    command -nargs=+ -complete=file -bar Ag silent! grep! <args>|cwindow|redraw!
+    nnoremap \ :Ag<SPACE>
+  endif
 endif
 
 " Make it obvious where 80 characters is


croaky:
Looks great! Thanks, @geoffharcourt

geoffharcourt:
@croaky we could also use `Ag.vim` or `Ack.vim` and use their command. Those plugins have the added benefit of allowing the results to be opened in splits, etc. Thoughts?

erickbrower:
Very cool, thanks guys! 👍

croaky:
Looks like `Ag.vim` is deprecated. I don't feel strongly about using `Ack.vim` one way or the other. Maybe a little confusing to have an `ack` reference if we go that route? Would want to maybe document that in the README next to Ag references so it's a little more explicit if we go that route.

I've been using these proposed lines since at least that blog post (2.5 years) and it's been working well for me.

geoffharcourt:
That works for me. I've confirmed this won't overwrite plugin-supplied `:Ag` commands. Thanks for the review.

benoittgt:
What about adding `!` to :Ag to avoid opening the first result ?

PR #498, opened 2016-07-25
[Vim] Don't open first result by default with \

benoittgt:
By default Ag.vim open the first result (https://github.com/thoughtbot/dotfiles/commit/e51d8b73f0d8ab99d1a12efdfbcd6dd287bec15c). To avoid it you can add a bang.

Their is a PR to make it default but it's not merge : https://github.com/rking/ag.vim/issues/119

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -75,7 +75,7 @@ if executable('ag')
 
   if !exists(":Ag")
     command -nargs=+ -complete=file -bar Ag silent! grep! <args>|cwindow|redraw!
-    nnoremap \ :Ag<SPACE>
+    nnoremap \ :Ag!<SPACE>
   endif
 endif
 


geoffharcourt:
@benoittgt do you have [ag.vim](https://github.com/rking/ag.vim) installed? This mapping doesn't use the `:Ag` command defined by ag.vim, it's using one we defined in the line above. Without an additional plugin, that mapping triggers an error:

```
E477: No ! allowed
```

benoittgt:
My bad. Sorry for the wrong PR.

geoffharcourt:
@benoittgt no problem! I'm glad there are people who aren't using our dotfiles directly who are watching the changes and making suggestions.

PR #500, opened 2016-09-28
Add pre-push git hook

bradurani:
Adds a pre-push git hook that calls the local pre-push git hook

I discovered this is useful when trying to add a pre-push hook that warns me when I push to `master`. I guess it's overkill to add all possible git hooks, but this one at least I found a real world use case for.


geoffharcourt:
I was about to ask if you could format this more like the other hooks, but then I realized we've got two slightly different styles going on in the hooks. LGTM.

geoffharcourt:
Thanks for your contribution, @bradurani!

PR #501, opened 2016-10-31
Allow user to ignore some default vim plugins

geoffharcourt:
This change wraps calls to `vim-plug`'s `Plug` command in
`vimrc.bundles` in a new command that first checks if the user has
blacklisted the plugin in `g:plugins_to_ignore`.

Users can declare their plugin blacklist in `$HOME/.vimrc.ignorebundles`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,24 @@ Your `~/dotfiles-local/vimrc.local` might look like this:
     highlight NonText guibg=#060606
     highlight Folded  guibg=#0A0A0A guifg=#9090D0
 
+If you don't wish to install a vim plugin from the default set of vim plugins in
+`.vimrc.bundles`, you can ignore the plugin by calling it out with `UnPlug` in
+your `~/.vimrc.bundles.local`.
+
+    " Don't install vim-scripts/tComment
+    UnPlug 'tComment'
+
+`UnPlug` can be used to install your own fork of a plugin or to install a shared
+plugin with different custom options.
+
+    " Only load vim-coffee-script if a Coffeescript buffer is created
+    UnPlug 'vim-coffee-script'
+    Plug 'kchmck/vim-coffee-script', { 'for': 'coffee' }
+
+    " Use a personal fork of vim-run-interactive
+    UnPlug 'vim-run-interactive'
+    Plug '$HOME/plugins/vim-run-interactive'
+
 To extend your `git` hooks, create executable scripts in
 `~/dotfiles-local/git_template.local/hooks/*` files.
 
--- a/vimrc.bundles
+++ b/vimrc.bundles
@@ -2,6 +2,14 @@ if &compatible
   set nocompatible
 end
 
+" Remove declared plugins
+function! s:UnPlug(plug_name)
+  if has_key(g:plugs, a:plug_name)
+    call remove(g:plugs, a:plug_name)
+  endif
+endfunction
+command!  -nargs=1 UnPlug call s:UnPlug(<args>)
+
 " Shim command and function to allow migration from Vundle to vim-plug.
 function! VundleToPlug(vundle_command, arg, ...)
   echom "You are using Vundle's `".a:vundle_command."` command to declare plugins. Dotfiles now uses vim-plug for plugin management. Please rename uses of `".a:vundle_command."` to `Plug`. Plugin was '".a:arg."'."


christoomey:
Rather than exposing a new command and config file, I wonder if we can just add a single new command that modifies VimPlug's configuration? Specifically, VimPlug uses a global dictionary to track plugins, namely `g:plugs`. We could introduce `UnPlug` which would remove the entry for the specified plugin:

```vim
" note: I haven't tested this code, but I'm pretty sure it's close to what we'd need
function! s:UnPlug(plug_name)
  if has_key(g:plugs, a:plug_name)
    call remove(g:plugs, a:plug_name)
  endif
endfunction
command!  -nargs=1 UnPlug call s:PlugOrIgnore(<args>)
```

The potential benefits to this approach are simplicity & backwards compatibility as we're only adding a single new command. In addition, with this approach we don't have to remember to use a custom `Plug` command variant.

The potential drawback is relying on the internals of VimPlug. I feel like this is likely safe due to how much of VimPlug relies on this global dictionary itself, but worth considering.

geoffharcourt:
I like this approach. I'll give this a try and see if there's any other things we haven't considered. Thanks!

geoffharcourt:
@christoomey I tweaked your suggestion in d9464f6. While this approach relies on vim-plug's internals, I think it's less intrusive than my original suggestion.

README.md
```diff
@@ -89,6 +89,17 @@ Your `~/.vimrc.local` might look like this:
     highlight NonText guibg=#060606
     highlight Folded  guibg=#0A0A0A guifg=#9090D0
 
+If you don't wish to install a vim plugin from the default set of vim plugins in
+`.vimrc.bundles`, you can ignore the plugin by calling it out with `UnPlug` in
+your `~/.vimrc.bundles.local`.
+blacklist in `~/.vimrc.ignorebundles`:
```
teoljungberg: Since we're relying on `vim-plug`'s internals, I think we can safely remove this file from the documentation.

README.md
```diff
@@ -89,6 +89,17 @@ Your `~/.vimrc.local` might look like this:
...
+
+    " Don't install vim-scripts/tComment
+    UnPlug 'tComment'
+
+`UnPlug` can be used to install your own fork of a plugin or to install a shared
+plugin with different custom options.
```
christoomey: I think it could be useful to provide an example for this. Maybe two lines, one with `UnPlug` and one with `Plug` adding it back with specific options.

geoffharcourt:
Gah, @teoljungberg & @christoomey I cleaned up that documentation editing error and added some custom setting examples in 07c19f556bbd0f86a09f345cdc10996a439da47d.

README.md
```diff
@@ -89,6 +89,24 @@ Your `~/.vimrc.local` might look like this:
...
     highlight Folded  guibg=#0A0A0A guifg=#9090D0
 
+If you don't wish to install a vim plugin from the default set of vim plugins in
+`.vimrc.bundles`, you can ignore the plugin by calling it out with `UnPlug` in
+your `~/.vimrc.bundles.local`.
+
+    " Don't install vim-scripts/tComment
+    UnPlug 'tComment'
+
+`UnPlug` can be used to install your own fork of a plugin or to install a shared
```
christoomey: In reading your code here I'm realizing that the plugins are stored in the `g:plugs` dictionary as just the plugin name, not `author/plugin-name` which I would have initially expected. This was a surprise to me. Do you think it's worth highlighting this in the summary?

vimrc.bundles
```diff
@@ -2,6 +2,14 @@ if &compatible
   set nocompatible
 end
 
+" Remove declared plugins
+function! s:UnPlug(plug_name)
+  if has_key(g:plugs, a:plug_name)
```
christoomey: Following on with my comment about re: `UnPlug` needing to work with just the pluging name, not `author/plugin-name`, do we think it would be feasible / useful to either warn or handle the case where a user tries to unplug with `author/plugin-name` argument?

``` vim
if match(a:plug_name, '\v^[^\/]+\/[^\/]+$')
  let plug_name_without_author = split(a:plug_name, '/')[1]
endif
```

README.md
teoljungberg: Ditto.

vimrc.bundles
teoljungberg: Agreed.

README.md
```diff
@@ -92,14 +92,21 @@ Your `~/.vimrc.local` might look like this:
...
 
     " Don't install vim-scripts/tComment
     UnPlug 'tComment'
 
 `UnPlug` can be used to install your own fork of a plugin or to install a shared
 plugin with different custom options.
 
+    " Only load vim-coffee-script if a Coffeescript buffer is created
+    UnPlug 'vim-coffee-script'
+    Plug 'kchmck/vim-coffee-script', { 'for': 'coffee' }
```
seanpdoyle: This seems like a good suggestion (whether we're `Unplug`ing or not).

Should take this approach for some of our the Plugs we include by default?

geoffharcourt: I've been doing this for several of the plugins I load from my personal config and the mechanism seems to work without problems.

The plugin that really affects startup time is `rails.vim`, and there's good reason to have that load even if you're not opening a Ruby buffer. We could do this for some of the plugins, but I don't think we'd get a huge performance benefit.

geoffharcourt: Update: I added lazy loading to the non-rails.vim plugins that seemed like the most likely candidates for load time improvement (only used for specific filetypes, didn't need to be preloaded like rails.vim): `vim-coffee-script`, `vim-go`, `vim-ruby`, and `vim-rake`, `vim-slim`. I wasn't able to shave any noticeable startup time when starting Neovim or Vim by lazy loading. 😞

MattMSumner:
@geoffharcourt this looks like exactly what I need right now. 👍

geoffharcourt:
@MattMSumner I'm hoping to clean this up on investment soon.

anhari:
Just wanted to say that this will be great for those that want to use an asynchronous linting plugin in place of syntastic.  Looking forward to this!  :+1:

derekprior:
@anhari I agree, though it's not too hard to do now (at the cost of loading, but not executing, two plugins). See: https://robots.thoughtbot.com/my-life-with-neovim#asynchronous-checkers

vimrc.bundles
geoffharcourt: Updated in ca4f857. I had to add a condition to `match` or it failed.

README.md
geoffharcourt: Updated in ca4f857.

geoffharcourt:
@christoomey I had to fiddle with the pattern you provided and include a `> 0` condition on `match to make this work on my local machine. Can you review my changes so we can (finally) close this out? Thank you for all the feedback, this made the change much better.

croaky:
Very nice. I came interested in finding solutions to what people are doing with Syntastic vs ALE.

christoomey:
@geoffharcourt This LGTM! 🚢

PR #503, opened 2016-12-01
Add `editorconfig-vim`

iwz:
Many projects are now using a `.editorconfig` file to share coding styles across editors. This adds support to vim for reading a project's `.editorconfig`.

See: http://editorconfig.org/

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -150,6 +150,9 @@ set complete+=kspell
 " Always use vertical diffs
 set diffopt+=vertical
 
+" Disable editconfig for fugitive
+let g:EditorConfig_exclude_patterns = ['fugitive://.*']
+
 " Local config
 if filereadable($HOME . "/.vimrc.local")
   source ~/.vimrc.local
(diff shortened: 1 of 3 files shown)


vimrc
```diff
@@ -150,6 +150,9 @@ set complete+=kspell
 " Always use vertical diffs
 set diffopt+=vertical
 
+" Disable editconfig for fugitive
+let g:EditorConfig_exclude_patterns = ['fugitive://.*']
```
geoffharcourt: Can you put this line into `vim/plugins/editorconfig.vim`?

iwz:
@geoffharcourt done

geoffharcourt:
@iwz this looks quite interesting. I was doing an audit of the plugin and I think this might be the first time we've bundled a plugin by default that uses Python code. I want to mull this over a little.

geoffharcourt:
Hey @iwz, adding expanded my Vim startup time by 40%. Given the bulk and the Python requirement I think this is probably a better fit as an opt-in for users in their personal overrides.

PR #508, opened 2017-01-12
Add git dirty status to zsh prompt

anhari:
I thought it would be nice to have a git status indicator included in the dotfiles by default.

- Displays a red exclamation mark when unstaged changes exist
- Displays a blue exclamation mark when staged changes exist
- Display a cyan asterisk when unpushed commits exist

Both exclamation marks representing uncommitted changes will take precedent over the asterisk.

git-status-zsh-prompt

### DIFF
--- a/zsh/configs/prompt.zsh
+++ b/zsh/configs/prompt.zsh
@@ -1,9 +1,28 @@
 # modify the prompt to contain git branch name if applicable
 git_prompt_info() {
-  current_branch=$(git current-branch 2> /dev/null)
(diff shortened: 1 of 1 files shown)


geoffharcourt:
Hi @anhari, thanks for your contribution to dotfiles.

How much git info in prompts appears to widely differ amongst users. (I use a system similar to yours, but I've noticed in pairing that it's not universally liked). 

I think this idea works better as an opt-in rather than a global default for dotfiles users.

PR #509, opened 2017-01-15
Move aliases below local config

dtt101:
Enables overriding of aliases that environment variables


geoffharcourt:
LGTM. Thanks for your contribution.

PR #510, opened 2017-01-19
Explicitly specify no input from terminal in post-up hook

geoffharcourt:
This change prevents the `post-up` hook from breaking the terminal for
the rest of the session when running `rcup` with a `Vim: Warning: Input is not from a terminal`.

Since the session is completely automated, we don't need input from the terminal once Vim
launches.

### DIFF
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -6,7 +6,7 @@ if [ ! -e "$HOME"/.vim/autoload/plug.vim ]; then
   curl -fLo "$HOME"/.vim/autoload/plug.vim --create-dirs \
       https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 fi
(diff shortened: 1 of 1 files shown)


geoffharcourt:
This solves my problem and I think won't have adverse effects for other users, but I wanted to review the approach to be sure this isn't a bad idea. cc: @thoughtbot/shell

hooks/post-up
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 
-touch "$HOME"/.psqlrc.local
+# touch "$HOME"/.psqlrc.local
```
mike-burns: Is this intentional?

geoffharcourt: Ack, no.

geoffharcourt: Reverted that diff in 6d79d65df4975931e670122e99d03bc5fedeb2a0.

Issue #511, opened 2017-01-23
README: Make your own customizations


rootulp:
I think the [Make your own customizations](https://github.com/thoughtbot/dotfiles#make-your-own-customizations) section could be made clearer by instructing users to use `~/dotfiles-local`. Currently, the explanation for `rcup` includes the following bullet point: 

> Give precedence to personal overrides which by default are placed in `~/dotfiles-local`

but the make your own customizations does not instruct users to place their local customizations in `~/dotfiles-local`. Instead it reads:

> Put your customizations in dotfiles appended with `.local`:
> - ~/.aliases.local
> - ~/.git_template.local/*
> - ...

I think this section should read:

> Create a directory for your personal customizations: `mkdir ~/dotfiles-local`
> Put your customizations in `~/dotfiles-local` appended with `.local`: 
> - ~/dotfiles-local/aliases.local
> - ~/dotfiles-local/git_template.local/*
> - ...

This makes it easier for users to version control their personal customizations in `dotfiles-local`
I'd be happy to make the changes if this seems sensible.

anhari:
I'd include the rcup command to sync those as well: `rcup -d ~/dotfiles-local -v`

geoffharcourt:
@anhari I think the existing [`.rcrc`](https://github.com/thoughtbot/dotfiles/blob/master/rcrc) handles that for you.

geoffharcourt:
In #513 the documentation was updated. Closing!

anhari:
So I recently had the idea of spinning up a `dotfiles-local` template repo that could be forked by new users as well.  I've actually switched an entire team to use this set up and the only painstaking process is manually spinning up a repo and creating local files for customizations.  What do you think of thoughtbot hosting a forkable `dotfiles-local` repo for new users?

PR #513, opened 2017-01-31
README: Make customizations in ~/dotfiles-local

rootulp:
- Change all references of the form `~/.config.local` to `~/dotfiles-local/config.local`
- References #511

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,26 +53,30 @@ any new files in the repository.
 Make your own customizations
 ----------------------------
 
-Put your customizations in dotfiles appended with `.local`:
+Create a directory for your personal customizations: 
 
-* `~/.aliases.local`
-* `~/.git_template.local/*`
-* `~/.gitconfig.local`
-* `~/.gvimrc.local`
-* `~/.psqlrc.local` (we supply a blank `.psqlrc.local` to prevent `psql` from
+    mkdir ~/dotfiles-local
+
+Put your customizations in `~/dotfiles-local` appended with `.local`:
+
+* `~/dotfiles-local/aliases.local`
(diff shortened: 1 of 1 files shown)


rootulp:
@geoffharcourt may this be merged?

README.md
```diff
@@ -53,26 +53,30 @@ any new files in the repository.
...
 
-Put your customizations in dotfiles appended with `.local`:
+Create a directory for your personal customizations: 
 
-* `~/.aliases.local`
-* `~/.git_template.local/*`
-* `~/.gitconfig.local`
-* `~/.gvimrc.local`
-* `~/.psqlrc.local` (we supply a blank `.psqlrc.local` to prevent `psql` from
+    mkdir ~/dotfiles-local
```
croaky: I don't personally use `~/dotfiles-local` as my directory. I use:

```
~/src/github.com/croaky/dotfiles
~/src/github.com/thoughtbot/dotfiles
```

https://robots.thoughtbot.com/manage-team-and-personal-dotfiles-together-with-rcm

rootulp: I use `~/dotfiles-local` because the default [rcrc](https://github.com/thoughtbot/dotfiles/blob/master/rcrc#L2) file uses it. If you think some users will put their personal customizations in a different folder then we can perhaps include a note on how to use rcm with another folder

croaky: Nice! Maybe mention the `rcrc` configuration option right before or after the first time `~/dotfiles-local` is referenced in this README? Then keep the rest as-is?

rootulp: Added!

geoffharcourt:
Squashed and merged as 09d54d4. 

Thanks for contributing to dotfiles!

PR #514, opened 2017-02-17
Add support for Elixir and EEx syntax highlighting

SViccari:


vimrc.bundles
```diff
@@ -47,6 +47,7 @@ Plug 'tpope/vim-rhubarb'
 Plug 'tpope/vim-surround'
 Plug 'vim-ruby/vim-ruby'
 Plug 'vim-scripts/tComment'
+Plug 'elixir-lang/vim-elixir'
```
geoffharcourt: I think it's probably time to add this! Can you `:sort` it with the other plugins?

SViccari: absolutely 👍

PR #518, opened 2017-04-26
Remove git churn command

croaky:
I added this four years ago but haven't been using it.
Not common enough to be included in company-wide dotfiles?
Can move to personal dotfiles if desired.


geoffharcourt:
This is a thing I've always wanted to use, but haven't managed to get it into my workflow rhythm.

PR #519, opened 2017-04-27
Remove gvimrc

croaky:
This originally came from jferris/dotfiles,
the parent of thoughtbot/dotifles.

I'm not a Vim GUI user.
Are others at thoughtbot also shell Vim users?


geoffharcourt:
I suspect there's still a lot of Macvim folks using dotfiles.

jferris:
The `gvimrc` doesn't have any best practice or "you obviously need this" kind of things, so I think it would be fine for MacVim users to make their own.

geoffharcourt:
Thanks!

PR #521, opened 2017-04-28
Run `reset` in post-up hook to fix terminal issue

### LINKED ISSUE #516: Running rcup results in strange terminal state requiring reset
Running `env RCRC=$HOME/dotfiles/rcrc rcup` after successful `brew install rcm` results in Vim error and leaves terminal in a strange state that requires a reset.

Error:
```
Vim: Warning: Input is not from a terminal
Vim: Caught deadly signal ABRT
Vim: Finished.
```

Formatting in terminal no longer has line breaks and requires reset to appear as intended.

geoffharcourt:
Some tmux users have reported that when running `rcup` that during the
post-up hook Vim would issue errors about input not coming from a
terminal and then the terminal would look and behave unexpectedly.

Running `reset` after the Vim plugin install step in `post-up` appears
to resolve the issue.

Fix #516.

### DIFF
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -8,6 +8,8 @@ if [ ! -e "$HOME"/.vim/autoload/plug.vim ]; then
 fi
 vim -u "$HOME"/.vimrc.bundles +PlugInstall +PlugClean! +qa
 
+reset -Q
+
 # detect old OS X broken /etc/zshenv and suggest rename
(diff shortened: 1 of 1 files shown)


mike-burns:
Seems right to me, but someone should probably confirm that it fixes their bug.

Discussed elsewhere: `-Q`.

devinrm:
I can confirm `reset -Q` fixes the bug and does not produce any output.

PR #522, opened 2017-05-01
Don't rely on Bash in post-up hook

### LINKED ISSUE #517: post-up hook is bash specific and doesn't honor /usr/bin/env
It is hard coded to `/bin/bash`

This is literally the only thing keeping things from playing nicely in FreeBSD and systems where bash isn't installed by default.

I have a fix but I'd like to figure out the route:

If you use`/usr/bin/env` in the `post-up hook`, it'll still require /bin/bash to be installed. However this is literally the only file that needs it. Can someone figure out if this would work with a zsh shebang line? Make sure to use `/usr/bin/env zsh` because many systems install their shells in `/usr/local/bin` or `/usr/bin`

geoffharcourt:
In 16219a31db5ca1f8c766d3c6066399950ff7f2c3 we implemented a post-up
hook that checked for a problematic `/etc/zshenv` file. OSX/macOS has
changed the file responsible for executing `path_helper` and loading the
system paths multiple times, and this script change issued a warning to
users if `path_helper` might be called in such a way that their paths
could be unexpectedly reordered.

The script used colored output to highlight this problem for the user,
and relied on Bash's script path introspection (through `$BASH_SOURCE`)
to print the location of the `post-up` hook to assist the user in
troubleshooting and understanding where our warning came from.

This change removes the hook's reliance on GNU Bash in favor of `sh`. It
uses the POSIX-compliant `$0` to determine the hook location (this
method is less robust than Bash's `$BASH_SOURCE`, but is available in
all POSIX shells) and removes colored output in the warning to be
compatible with shells that don't support colored output.

Fix #517.

### DIFF
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 touch "$HOME"/.psqlrc.local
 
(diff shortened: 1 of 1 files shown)


geoffharcourt:
I'd like to solicit input on this before merging. cc: @thoughtbot/shell @mislav

I couldn't get the colored output to work with dash on macOS, but if there's a way to retain the color in the warning I would be very happy to pursue that.

I did a bunch of reading about `$0` and its flaws/drawbacks, but it seems that this approach is acceptable for our use case because of the way in which `post-up` gets invoked.

geoffharcourt:
One more note: We're about 18 months since the last time a macOS update made a big change to the zsh config.

hooks/post-up
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
```
teoljungberg: We are consistently using `/bin/sh` as our `sh` path. I haven't seen anything noting that the `sh` shell should live elsewhere.

mike-burns: +1 to this. We don't support old Solaris and their lack of `$(...)`.

geoffharcourt:
I'm sad to lose the colors, but I'm excited about making dotfiles usable by a slightly broader audience. Thanks for the feedback.

PR #523, opened 2017-05-07
Add JavaScript syntax highlighting

croaky:
It seems like JavaScript/ES6 is becoming more common for us to run into
in the wild compared to CoffeeScript
with the advent of Babel / transpiling.

Eventually, we'll probably want to remove kchmck/vim-coffee-script.
Users working on CoffeeScript applications would need to move that
plugin into their personal dotfiles.


geoffharcourt:
I think adding the JS plugin makes a lot of sense. However, I still end up working on client projects that have existing Coffeescript fairly frequently. It makes sense for us to prefer ES6, but given that there's still a lot of Coffeescript in the wild (and particularly in Rails apps), I think we should keep the vim-coffeescript plugin for now. We could lazy-load it with vim-plug if we're concerned about VIm performance.

MattMSumner:
I would be ok with removing this plugin. I haven't looked at any coffee script in over a year.

croaky:
Updated to only add JavaScript, not remove CoffeeScript.

PR #526, opened 2017-05-23
Skip compiled wordcode files (ie. zwc extensions)

jeckhart:
zcompiled files will cause errors for the zsh script as it starts.

### DIFF
--- a/zshrc
+++ b/zshrc
@@ -10,6 +10,7 @@ _load_settings() {
   if [ -d "$_dir" ]; then
     if [ -d "$_dir/pre" ]; then
       for config in "$_dir"/pre/**/*(N-.); do
+        if [ ${config:e} = "zwc" ] ; then continue ; done
         . $config
       done
     fi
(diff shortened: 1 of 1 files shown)


geoffharcourt:
cc: @thoughtbot/shell

mike-burns:
Oh that's wild.

How about merging it with the conditional one line up?

jeckhart:
Updated!

mike-burns:
LGTM.

geoffharcourt:
Merged! Thank you @jeckhart for your research and PR.

geoffharcourt:
@jeckhart my apologies, this commit made all of my zsh configs not work (they were never sourced). I reverted in 88618e5 until we can figure out what the problem is an then re-implement.

jeckhart:
My bad, the problem stems from the single bracket comparison on [line 27](https://github.com/jeckhart/dotfiles-1/blob/skip-zwc-files/zshrc#L27). It needs a double bracket (which is built into zsh and doesn't have to abide by the legacy sh bracket behaviors). It was in my local copy when testing but missed when I merged it. I'll submit a new PR.

PR #527, opened 2017-05-25
Skip compiled wordcode files (ie. zwc extensions)

jeckhart:
Revises my changes from PR #526 after proper testing.


geoffharcourt:
Merged as f4d509f. Thanks for the revision!

PR #529, opened 2017-06-06
Add potential fix for Mac OS X copy-paste.

StevenXL:
I really think alluding to a potential solution would save others a ton of time. Almost all Mac users would find this useful - there's a lot of programmers that use Macs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,13 @@ Shell aliases and scripts:
 * `tat` to attach to tmux session named the same as the current directory.
 * `v` for `$VISUAL`.
 
+Mac OS X
+------
+
+If you are having trouble with copying from / to the Mac OS X system clipboard,
+the [following post][vimcopy] may be able to help.
+
+
 Thanks
 ------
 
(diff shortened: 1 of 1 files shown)


gohanlon:
I tend to agree, and maybe alluding to the solution is a good approach as I think thoughtbot/dotfiles themselves are purposefully OS agnostic or at least currently lack specific macOS and Linux tidbits.

My own [dotfiles-local](https://github.com/gohanlon/dotfiles-local) offer a [solution](https://github.com/gohanlon/dotfiles-local/commit/7d704d96b6469806048ec89da87948f9fb2a2e0c) that's very similar to the Evert Pot blog post you referenced. My approach is shell agnostic and uses Tmux's current configuration syntax (which changed in Tmux 2.4, I think).

On a more minor note, I personally don't configure Vim with `set clipboard=unnamed` (mentioned in the blog post) because I don't want all the Vim commands that manipulate the unnamed register to wipe my system clipboard. Instead, I omit that setting and just use Vim's quoteplus register to deliberately work with the system clipboard, e.g. copy to system clipboard with `"+yy` and paste from it with `"+p`.

Hopefully the technique I'm using will be valid for a long time, but it will break again at some point due to changes to either Vim, Tmux, or macOS. I'd love to have some canonical place to document the solution as it changes over time. Just not sure where that'd be.

StevenXL:
@gohanlon I know exactly what you mean, and my setup is similar in that i don't use `set clipboard=unnamed`. I'm going to take a look at your dotfiles to see what you mean in terms of Tmux's current configuration syntax. Thank you.

PR #532, opened 2017-07-07
fix #524, update plugins

### LINKED ISSUE #524: Problems with UnPlug?
Think I'm doing something wrong here:

```
dotfiles ale % v laptop.local
Error detected while processing function plug#end:
line   15:
E716: Key not present in Dictionary: syntastic
E15: Invalid expression: g:plugs[name]
Error detected while processing function plug#end[71]..<SNR>7_reorg_rtp:
line   13:
E716: Key not present in Dictionary: syntastic
E116: Invalid arguments for function s:rtp(g:plugs[v:val])
Error detected while processing /Users/croaky/.vimrc.local:
line   63:
E185: Cannot find color scheme 'github'
Press ENTER or type command to continue
```

The above happens when I add `UnPlug 'syntastic'` to the top of my `vimrc.bundles.local`. I can fix it by commenting it out:

https://github.com/croaky/dotfiles/compare/ale?expand=1

I'm guessing it's because I have my personal dotfiles listed first with https://github.com/thoughtbot/rcm

`DOTFILES_DIRS="$HOME/src/github.com/croaky/dotfiles $HOME/src/github.com/thoughtbot/dotfiles"`

If I list thoughtbot/dotfiles last, there's no way to use `UnPlug`, right?

floesche:
* updating and upgrading existing plugins seems consistent with the idea of [rcup](http://thoughtbot.github.io/rcm/rcup.1.html)
* if .vim/autoload/plug.vim already exists, then use PlugUpgrade to get the newest version. This should also fix #524.
* change from PlugInstall to PlugUpdate which doesn't just install new plugins but also updates existing ones.

### DIFF
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -2,11 +2,11 @@
 
 touch "$HOME"/.psqlrc.local
 
-if [ ! -e "$HOME"/.vim/autoload/plug.vim ]; then
+if [ -e "$HOME"/.vim/autoload/plug.vim ]; then
+  vim -E -s +PlugUpgrade +qa
+else
   curl -fLo "$HOME"/.vim/autoload/plug.vim --create-dirs \
       https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
-else
-  vim -E -s +PlugUpgrade +qa
 fi
 vim -u "$HOME"/.vimrc.bundles +PlugUpdate +PlugClean! +qa
 
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -5,8 +5,10 @@ touch "$HOME"/.psqlrc.local
 if [ ! -e "$HOME"/.vim/autoload/plug.vim ]; then
   curl -fLo "$HOME"/.vim/autoload/plug.vim --create-dirs \
       https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
+else
+  vim -E -s +PlugUpgrade +qa
 fi
-vim -u "$HOME"/.vimrc.bundles +PlugInstall +PlugClean! +qa
+vim -u "$HOME"/.vimrc.bundles +PlugUpdate +PlugClean! +qa
 
 reset -Q
 


gohanlon:
This seems good to me. I think this is the behavior that a new user would expect, and I don‘t see any disadvantages. (Are there any?)

Personally, I run `rcup` a lot more than `:PlugUpdate`, and I don't micro-manage plugin updates. This change would result in my plugins being kept more up-to-date and eliminate a separate step (that I run much less often).

One tiny nit: I'd prefer that the `if` condition were invested, so it'd read as `if exists; else` rather than `if not exists; else`, that is:

``` sh
if [ -e "$HOME"/.vim/autoload/plug.vim ]; then
  vim -E -s +PlugUpgrade +qa
else
  curl -fLo "$HOME"/.vim/autoload/plug.vim --create-dirs \
fi
```

Thanks!

floesche:
According to [vim.plug's documentation](https://github.com/junegunn/vim-plug/blob/master/README.md#commands) `PlugUpdate` installs new plugins just like `PlugInstall`, but in addition updates existing ones. 

I am not 100% sure how `plug.vim` determines what is going to be updated. From my understanding of the code `PlugUpdate` and `PlugInstall` both call the same function and rely on `g:plugs`, in which case there would be no disadvantage. Should the update also consider the plugin directory and update everything in there, then it might trigger an unnecessary download before removing the unwanted plugin through `PlugClean!`. That's a potential disadvantage if I overlooked something in the `plug.vim` code.

Thanks for the suggestion, in 05fc3d4cca69dca4b3f79ba15372c63d5e37e11d I changed the `if` to the better readable order you suggested.

geoffharcourt:
I update my own plugins fairly aggressively, but I am concerned that this might be a surprise to users.

floesche:
I understand this reservation in regards to using `PlugUpdate` instead of `PlugInstall`, even though it was a surprise to me that `rcup` does not update existing plugins, only install new ones. Maybe this could at least be clarified in the README / documentation?

In regards to `plug.vim` itself I think it's a slightly different argument: in the current version it is pulled in during the first run of `rcup` which results in different and quite random versions of `plug.vim` installed across different machines, based on the exact time when the dotfiles are initialised. This was presumably the cause of #524 for at least two users, who didn't seem to expect the exhibited behaviour. For my part, I was expecting that a `git pull` followed by a `rcup` would bring all the elements provided by https://github.com/thoughtbot/dotfiles/ up to the latest version.

If you think that `PlugUpdate` is too aggressive, but can follow my argument to the surprising difference in `plug.vim` across different machines, I'd be happy to provide a patch with only that part.

croaky:
I've tried this on my machine and it appears to work as advertised, seems like it addresses my original #524.

Just about everything else in my `rcup` and https://github.com/thoughtbot/laptop setup is intended to be idempotent, run whenever to update to the latest system dependencies.

Something like this feels necessary because the Plug install approach is to `curl` a raw file from GitHub. Would be nice to be able to use a package manager like NPM or something kind of like the Atom / VS Code communities but this seems like what we have with Vim? I'm good with it.

geoffharcourt:
Squashed and merged as d6014c3. Thanks for your contribution!

PR #533, opened 2017-08-04
Add `bin/git-trust-bin` command

seanpdoyle:
To [append a project's local `bin/` directory to `$PATH`][dotfiles],
the dotfiles check for the existence of a special directory: `git/.safe`.

This commit is inspired by [thoughtbot/suspenders#837][#837]:

> The idea behind the .git/safe is that I run it manually and explicitly
> after I trust the directoy and the team behind the project.

> Creating it as part of the setup script defeats the point of that. Doing
> so without explcitly telling the user comes across as sneaky, perhaps.
> (I do expect all devs to read setup scripts before running them -- but I
> also know that they do not!)

> Switch this script to inform the user that they have no `.git/safe`
> directory but might like to make one.

Since suspenders may no longer mark projects safe by default, this
commit introduces the `git safe` command to do so.

[dotfiles]: https://github.com/thoughtbot/dotfiles/blob/af75a673b1d8465a1e8c55c33f4a79fe3f5dc3c7/zsh/configs/post/path.zsh#L9-L10
[#837]: https://github.com/thoughtbot/suspenders/pull/837

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ any new files in the repository.
 Make your own customizations
 ----------------------------
 
-Create a directory for your personal customizations: 
+Create a directory for your personal customizations:
 
     mkdir ~/dotfiles-local
 
@@ -209,6 +209,7 @@ configuration:
 * Adds `post-{checkout,commit,merge}` hooks to re-index your ctags.
 * Adds `pre-commit` and `prepare-commit-msg` stubs that delegate to your local
   config.
+* Adds `trust-bin` alias to append a project's `bin/` directory to `$PATH`.
 
 [Ruby](https://www.ruby-lang.org/en/) configuration:
 
--- a/bin/git-trust-bin
+++ b/bin/git-trust-bin
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+mkdir -p .git/safe


derekprior:
I have a similar command, `git trust` which expresses more of what I'm doing. Perhaps being more explicit: `git trust-bin` would be better for a shared project?

gylaz:
Do we really need an alias fro something simple as `mkdir -p .git/safe`? I'm not opposed to it, but I've never felt the pain of typing `mkdir .git/safe` before. What's the chance that someone will know to use this new command/alias rather than creating the dir?

croaky:
Looks like the `thoughtbot/suspenders#837` link is broken?

seanpdoyle:
Since we used to [rely on bin/setup](https://github.com/thoughtbot/suspenders/pull/837), remembering (and even thinking about the significance) of `.git/safe` was never important.

The idea behind this PR is to introduce a command that shifts the burden of configuring a project _from_ the developer _to_ the dotfiles themselves. `.git/safe` is an arbitrary directory. It hasn't changed in the past, but if it needed to, we could modify this command to account for whatever changes were necessary.

It isn't mentioned anywhere in the README, and [the only current mention](https://github.com/thoughtbot/dotfiles/search?utf8=%E2%9C%93&q=git%2Fsafe&type=) is buried in a code comment in a post-hook.

As an alternative to introducing a new command could be to improve the current documentation around this. @croaky suggested that we write a ["be-all, end-all" blog post on the topic](https://github.com/thoughtbot/suspenders/pull/837#discussion_r131438836) that we could link to in the README.

geoffharcourt:
👍  @seanpdoyle!

PR #534, opened 2017-08-04
Use ASDF for version managment

derekprior:
ASDF has plugins for Ruby, Node, Elixir, Haskell, Scala, Go, etc.
Instead of finding and installing a new tool each time we have need for
a version manager in a new tools, we can use the same tool we already
know, adding a new plugin.

This change is the result of a successful [research card][1] and has a
corresponding change to [laptop][2].

[1]: https://trello.com/c/MVbfz8e5/676-asdf-for-tool-management
[2]: https://github.com/thoughtbot/laptop/pull/502

### DIFF
--- a/zsh/configs/post/path.zsh
+++ b/zsh/configs/post/path.zsh
@@ -5,6 +5,14 @@ PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 if [ -d "$HOME/.asdf" ]; then
   . $HOME/.asdf/asdf.sh
 elif command -v rbenv >/dev/null; then
+  if [ -z $SILENCE_RBENV_DEPRECATION ]; then
+    echo "The thoughtbot dotfiles have deprecated the use of rbenv in favor"\
+         "of asdf (https://github.com/asdf-vm/asdf) and will be removed on or"\
+         "after December 8, 2017.  Migrate to asdf or export"\
+         "SILENCE_RBENV_DEPRECATION=1 in your local dotfiles to silence this"\
+         "deprecation."
+  fi
+
   eval "$(rbenv init - --no-rehash)"
 fi
 
--- a/zsh/configs/post/path.zsh
+++ b/zsh/configs/post/path.zsh
@@ -1,8 +1,10 @@
 # ensure dotfiles bin directory is loaded first
 PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
-# load rbenv if available
-if command -v rbenv >/dev/null; then
+# load ASDF, falling back to rbenv if not available
+if [ -d "$HOME/.asdf" ]; then
+  . $HOME/.asdf/asdf.sh
+elif command -v rbenv >/dev/null; then
   eval "$(rbenv init - --no-rehash)"
 fi
 
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ configuration:
 [Ruby](https://www.ruby-lang.org/en/) configuration:
 
 * Add trusted binstubs to the `PATH`.
-* Load the ASDF version manager, falling back to rbenv if not installed.
+* Load the ASDF version manager.
 
 Shell aliases and scripts:
 
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ configuration:
 [Ruby](https://www.ruby-lang.org/en/) configuration:
 
 * Add trusted binstubs to the `PATH`.
-* Load rbenv into the shell, adding shims onto our `PATH`.
+* Load the ASDF version manager, falling back to rbenv if not installed.
 
 Shell aliases and scripts:
 


zsh/configs/post/path.zsh
```diff
@@ -1,8 +1,10 @@
...
 PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
-# load rbenv if available
-if command -v rbenv >/dev/null; then
+# load ASDF, falling back to rbenv if not available
+if [ -d "$HOME/.asdf" ]; then
+  . $HOME/.asdf/asdf.sh
+elif command -v rbenv >/dev/null; then
   eval "$(rbenv init - --no-rehash)"
 fi
```
croaky: Everything else in this file is for PATH-specific stuff. Should the version management portion be extracted to a separate file? Looks like the load order matters and is currently alphabetically by file? https://github.com/thoughtbot/dotfiles/blob/master/zshrc

zsh/configs/post/path.zsh
```diff
@@ -1,8 +1,10 @@
 # ensure dotfiles bin directory is loaded first
 PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
-# load rbenv if available
-if command -v rbenv >/dev/null; then
+# load ASDF, falling back to rbenv if not available
+if [ -d "$HOME/.asdf" ]; then
+  . $HOME/.asdf/asdf.sh
```
croaky: `asdf`s README also suggests:

```
. $HOME/.asdf/completions/asdf.bash
```

Worth it?

https://github.com/asdf-vm/asdf/issues/97#issuecomment-271482373

gabebw: I would assume that `asdf.bash` only works with bash, not Zsh.

Interestingly, Zsh _can_ understand Bash completion by running this line:

```
autoload -U bashcompinit && bashcompinit
```

croaky: I tried adding this in different places in our `zsh` directory and couldn't find one that got past the same error described in 
https://github.com/asdf-vm/asdf/issues/97#issuecomment-271482373

derekprior: Completions and zsh initialization order are the perfect storm of my dotfile nightmares. I think we can move forward without it and tackle later if someone is sufficiently motivated.

zsh/configs/post/path.zsh
```diff
@@ -1,8 +1,10 @@
...
+elif command -v rbenv >/dev/null; then
   eval "$(rbenv init - --no-rehash)"
```
derekprior: Should we add an echo here to encourage folks to move to ASDF or override in their local dotfiles? How long do we keep this fallback?

geoffharcourt: 3 months? I'm thinking the same thing about our Vundle -> vim-plug migration. Anyone waiting longer than that to re-sync can take on the burden of resolving conflicts.

derekprior:
@geoffharcourt Added a deprecation so we don't have to keep this `rbenv` bit in there forever. I thought about other potentially less annoying ways to do this -- maybe writing a file when we warn so we can do it once/day at most or something, but that seems like a lot of machinery. thoughts?

geoffharcourt:
@derekprior I'd like to merge this next week. I need some time to look this over carefully given the impact of the change, and I won't be able to do that today. 

Maybe we can handle deprecations with GH milestones so that the due dates in GH give us a cue to remove the deprecation?

derekprior:
We could also merge without the deprecation and make an issue to revisit the deprecation at a later date. The harm in supporting both is the tiny bit of slowness introduced by the conditional, I suppose.

geoffharcourt:
🎉 

Awesome change, @derekprior! Thank you for doing all the legwork to make this work and to give users an upgrade path.

StevenXL:
Hi. First, thank you for this awesome project. It has saved me a ton of time.

I am trying to migrate over to `asdf`, but I am having issues. Here are the steps I have taken:

1. Uninstalled `rbenv` via `brew uninstall rbenv`, and deleted `~/.rbenv` directory.
2. Followed [these directions](http://www.akitaonrails.com/2017/10/24/replacing-rvm-rbenv-nvm-etc-for-asdf) to install `asdf`.
3. Installed the ruby plugin for asdf, and installed ruby version 2.3.3 since that is what my project requires.

Note that I didn't add `echo -e '\n. $HOME/.asdf/asdf.sh' >> ~/.zshrc` since that is done by this project. 

Anway, when I run `asdf current ruby` in my rails project, I get the following output:

`2.3.3 (set by /Users/stevenleiva/Repos/<my_rails_project>/.tool-versions)`

And when I do `ruby --version` I get the following:

`ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin16]`

This all looks good to me. But then I run `bundle exec rails s` and I get this output:

`Your Ruby version is 2.4.1, but your Gemfile specified 2.3.3`

Version `2.4.1` is the version I have installed in `/usr/local/bin/ruby`. 

Sorry to post this here, but I couldn't find a better place. Any help is welcomed.

croaky:
@StevenXL What is your output for the following?

```sh
% which bundle
/Users/croaky/.asdf/shims/bundle
```

If it's not showing the shim for Bundler, try:

```sh
% gem install bundler
% asdf reshim ruby
```

StevenXL:
@croaky thank you. `which bundle` was pointing to `/usr/local....`. 

Followed your directions and now it points to the shim. Reinstalling gems. **Very** grateful for your help.

evercodes:
Add the following to ~/.zshrc
`for config_file ($HOME/.asdf/asdf.sh) source $config_file`

PR #535, opened 2017-08-06
Replace Syntastic with ALE

derekprior:
ALE is an Asynchronous Linting Engine which works with Vim 8 and NeoVim
to lint your code "as you type". It does so without blocking the main
thread by using the asynchronous features of both of those Vim versions.

A number of us in Boston have been using ALE for some time now with
great success. The asynchronous nature means linting does not block
other interactions with Vim and it's been nice to get linting feedback
more often than buffer write.

Many of us find the "as-you-type" linting to be performant enough to
actually be a distraction, so this PR tones that down some. With this
configuration, ALE will lint on cursor hold durations of 1 second and
when leaving insert mode in addition to buffer write, buffer read, and
file type change.

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -44,6 +44,14 @@ augroup vimrcEx
   autocmd BufRead,BufNewFile Appraisals set filetype=ruby
   autocmd BufRead,BufNewFile *.md set filetype=markdown
   autocmd BufRead,BufNewFile .{jscs,jshint,eslint}rc set filetype=json
+
+  " ALE linting events
+  set updatetime=1000
+  let g:ale_lint_on_text_changed = 0
+  autocmd CursorHold * call ale#Lint()
+  autocmd CursorHoldI * call ale#Lint()
+  autocmd InsertEnter * call ale#Lint()
+  autocmd InsertLeave * call ale#Lint()
 augroup END
 
 " When the type of shell script is /bin/sh, assume a POSIX-compatible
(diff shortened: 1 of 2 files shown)


christoomey:
I love ALE and I'm very much in favor of this switch. My only concern here is that this will break for users not yet on NeoVim / Vim 8+. I just tested locally with my system vim and it's a bunch of noisy ale-related warnings. I wonder if we might want to handle this edge case via any of:

1. Conditionally fall back to syntastic if `v:version < 800`?
2. Add a notice to the README about requiring NeoVim or Vim 8+
3. Printing an error when starting Vim

``` vim
if v:version < 800
  echoerr "The thoughtbot dotfiles require NeoVim or Vim 8 for async support"
endif
```

jferris:
I'm in favor of this pull request.

> Printing an error when starting Vim

I'm in favor of a pull request which does this.

jessieay:
I happened to update my dotfiles off of `master` just now and this update completely broke vim for me (was using my mac's built-in Vim version, was at ~7.4). 

Ugrading vim to 8.0 with homebrew fixed it, but I just wanted to echo @christoomey's suggestion to do something for people who are on get on Vim v8 because if I weren't such a great pull request sleuth this could have taken me much longer to figure out. 🦀

jferris:
@jessieay I think that will get better after #536 is merged, right?

jessieay:
@jferris you are right! did not see that one

PR #536, opened 2017-08-08
Require Vim 8 or NeoVim

derekprior:
Now that we're using `ALE`, we're making use of the async features
available in Vim8 and NeoVim. If you try to use these dotfiles on Vim 7,
you'll get a host of errors from ALE.

This change does two things:

1. It writes an error when starting in an unsupported vim version.
2. It skips loading ALE in those versions.

The thinking behind the second change is to leave vim in a more usable
state with minimal annoyance after the initial error.

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -12,6 +12,8 @@ set incsearch     " do incremental searching
 set laststatus=2  " Always display the status line
 set autowrite     " Automatically :write before running commands
 
+let g:has_async = v:version >= 800 || has('nvim')
+
 " Switch syntax highlighting on, when the terminal has colors
 " Also switch on highlighting the last used search pattern.
 if (&t_Co > 2 || has("gui_running")) && !exists("syntax_on")
@@ -46,7 +48,7 @@ augroup vimrcEx
   autocmd BufRead,BufNewFile .{jscs,jshint,eslint}rc set filetype=json
 
   " ALE linting events
-  if v:version >= 800
+  if g:has_async
     set updatetime=1000
     let g:ale_lint_on_text_changed = 0
     autocmd CursorHold * call ale#Lint()
@@ -55,7 +57,7 @@ augroup vimrcEx
     autocmd InsertLeave * call ale#Lint()
   else
     echoerr "The thoughtbot dotfiles require NeoVim or Vim 8"
-  end
+  endif
 augroup END
 
 " When the type of shell script is /bin/sh, assume a POSIX-compatible
--- a/vimrc
+++ b/vimrc
@@ -18,10 +18,6 @@ if (&t_Co > 2 || has("gui_running")) && !exists("syntax_on")
   syntax on
 endif
 
-if v:version < 800
-  echoerr "The thoughtbot dotfiles require NeoVim or Vim 8"
-endif
-
 if filereadable(expand("~/.vimrc.bundles"))
   source ~/.vimrc.bundles
 endif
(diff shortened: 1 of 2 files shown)


vimrc
```diff
@@ -46,12 +50,14 @@ augroup vimrcEx
   autocmd BufRead,BufNewFile .{jscs,jshint,eslint}rc set filetype=json
 
   " ALE linting events
-  set updatetime=1000
-  let g:ale_lint_on_text_changed = 0
-  autocmd CursorHold * call ale#Lint()
-  autocmd CursorHoldI * call ale#Lint()
-  autocmd InsertEnter * call ale#Lint()
-  autocmd InsertLeave * call ale#Lint()
+  if v:version >= 800
```
derekprior: I don't love repeating this conditional in both the bundles list and here. I'd feel better if there was a corresponding `feature-list` entry I could use like `has('async')` but I didn't see anything...

croaky: [if/else](http://learnvimscriptthehardway.stevelosh.com/chapters/21.html#else-and-elseif)?

derekprior: That would clean up the duplicated conditional in this file, for sure. Thanks.

I was thinking more about the duplication between this file and the bundles file. But at least it's only conceptually duplicated twice with that change...

gabebw: Should this also allow Neovim? `has('nvim')` will [check for it](https://github.com/neovim/neovim/issues/1931):

```
if v:version >= 800 || has('nvim')
```

derekprior: what does `v:version` return in NeoVim?

christoomey: My local (recent install) of neovim returns 704, so I think the `has('nvim')` addition makes sense. Since the conditional is a bit more complex now, what do you think of introducing a `function! s:has_async()` in the vimrc. I expect that the function will be available in the `vimrc.bundles` since it is sourced from the vimrc, although you might need to make it global rather than `s:` script local.

jferris: Here is the check that Ale itself uses: https://github.com/w0rp/ale/blob/master/plugin/ale.vim#L21

christoomey: I think we'd need the full conditional that lives in, which has a different branch for neovim:

``` vim
" A flag for detecting if the required features are set.
if has('nvim')
    let s:has_features = has('timers')
else
    " Check if Job and Channel functions are available, instead of the
    " features. This works better on old MacVim versions.
    let s:has_features = has('timers') && exists('*job_start') && exists('*ch_close_in')
endif
```

From [ale/ale.vim L16-22](https://github.com/w0rp/ale/blob/16772298ab555119223c50a7a8bf4800c436e2f9/plugin/ale.vim#L16-L22)

derekprior: I'm not really thrilled with duplicating all of that. Blargh.

Based on my reading of the script, Ale should already print a useful message. Is that the experience of folk who tried it on vim < 8?

Should we only guard the auto commands and then let the plugin do our warning for us? Or is it more clear if we explicitly say "the thoughtbot dotfiles require..."

jferris: This comment made it seem like it was confusing?

https://github.com/thoughtbot/dotfiles/pull/535#issuecomment-321065135

christoomey:
```

derekprior: Ah, that emit conflict warnings thing looks like it violates the spirit of this code: https://github.com/w0rp/ale/blob/16772298ab555119223c50a7a8bf4800c436e2f9/plugin/ale.vim#L24-L33

derekprior: I had to make it a global variable, but it's done.

croaky:
Niiice

PR #538, opened 2017-08-10
has_async is undefined when running rcup

rickpeyton:
@derekprior I pulled down master today and when I ran `rcup` it complained that `has_async` was undefined in `vimrc.bundles`

I simply copied the code that existed in `vimrc`

Is there a better way to do this that doesn't duplicate code?

### DIFF
--- a/vimrc.bundles
+++ b/vimrc.bundles
@@ -57,6 +57,8 @@ Plug 'tpope/vim-surround'
 Plug 'vim-ruby/vim-ruby'
 Plug 'vim-scripts/tComment'
 
+let g:has_async = v:version >= 800 || has('nvim')
+
 if g:has_async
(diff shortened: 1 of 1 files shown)


derekprior:
Ah. We forgot that on rcup we don't load vimrc.

derekprior:
There's probably something we can do but duplicating the conditional is fine for now. Maybe we should make the script variables instead of global if we are going to duplicate them anyway. 

@christoomey, do you have any ideas? I'll merge this when I get to the office otherwise.

derekprior:
@rickpeyton Thanks for reporting this. I think #539 fixes this without the duplication.

derekprior:
@rickpeyton Thanks for reporting this. I think #539 fixes this without the duplication.

PR #539, opened 2017-08-10
Move definition of `has_async`

derekprior:
We have an rcm `post-up` hook that installs plugins after running
`rcup`. It does this by directly loading the `.vimrc.bundles` file,
which means our regular `.vimrc` file is not sourced. This lead to
`has_async` being undefined.

By moving the definition into the `vimrc.bundles` we ensure it's
available when that `post-up` hook runs. It remains available when
needed in `vimrc` because we use it there *after* sourcing the bundles.

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -12,8 +12,6 @@ set incsearch     " do incremental searching
 set laststatus=2  " Always display the status line
 set autowrite     " Automatically :write before running commands
 
-let g:has_async = v:version >= 800 || has('nvim')
-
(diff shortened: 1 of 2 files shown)


christoomey:
This seems fine, but as an alternative, what if the `post-up` hook called into the main `.vimrc` directly? I expect the time added would be very minimal, and this means that the `.vimrc.bundles` only has one entry point rather than two?

derekprior:
I think we call into the bundles directly because the vimrc will contain configuration for plugins that aren't yet installed.

PR #541, opened 2017-08-19
Fix Personal Customization Path

ikanyu:
There are inconsistencies between the README and the personal
customization path. This commit is to fix the path to personal
customizations file as any customizations placed in `~/dotfiles-local`
will not be loaded correctly.

Lets take `zshrc.local` as an example. In README, it is stated that customizations have to be placed in `~/dotfiles-local/zshrc.local`. However, in the `zshrc`, the local config path points to `~/.zshrc.local` without the `/dotfiles-local` directory and has a `dot` file extension.

### DIFF
--- a/aliases
+++ b/aliases
@@ -16,4 +16,4 @@ alias s="rspec"
 alias path='echo $PATH | tr -s ":" "\n"'
 
 # Include custom aliases
-[[ -f ~/.aliases.local ]] && source ~/.aliases.local
+[[ -f ~/dotfiles-local/aliases.local ]] && source ~/dotfiles-local/aliases.local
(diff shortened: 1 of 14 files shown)


geoffharcourt:
Hi @ikanyu if you're using `rcm`, both the shared and local files will be symlinked under your home folder. It's not a requirement to have your personal customizations in `~/dotfiles-local` (although that's what lots of folks including me use). We don't want to force a hard-dependency on a specific files location.

If you want to use a different folder name, you can set that in your `.rcrc`.

ikanyu:
Thanks a lot for the comment and maintaining this dotfiles! TIL on the rcm!

PR #543, opened 2017-08-26
Add a specific "Update" section to the README

christoomey:
Thanks to rcm and the `post-up` hook file, the dotfiles do a great job
of keeping things up to date.

This change adds a bit of detail and specificity to the README around
updating, specifically clarifying the need to run `rcup` after pulling
to link new files and install new vim plugins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,12 +45,17 @@ configuration options:
 * Please configure the `rcrc` file if you'd like to make personal
   overrides in a different directory
 
-You can safely run `rcup` multiple times to update:
+
+Update
+------
+
+From time to time you should pull down any updates to these dotfiles, and run
 
     rcup
 
-You should run `rcup` after pulling a new version of the repository to symlink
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,12 +45,16 @@ configuration options:
...
+------
+
+From time to time you should pull down any updates to these dotfiles, and run
 
     rcup
 
-You should run `rcup` after pulling a new version of the repository to symlink
-any new files in the repository.
+to link any new files and install new vim plugins. You can safely run `rcup`
+multiple times, so update early and update often!
```
geoffharcourt: I think in the change here we are losing a little bit of the imperative. Can we add to this to state that pulling updates and _not_ running `rcup` may result in problems because there may be missing files or plugins?

christoomey:
Wording updated. Thoughts on this version @geoffharcourt?

PR #545, opened 2017-08-29
Add export to support Elixir/Erlang history

gylaz:
[Sourced from this](https://til.hashrocket.com/posts/is9yfvhdnp-enable-history-in-iex-through-erlang-otp-20-)


derekprior:
Nice! I remember reading about this but hadn't crossed the chasm to actually put it in my own dotfiles. I'd be thrilled to pull this from here!

dbernheisel:
Heck ya 👍

PR #547, opened 2017-09-26
Filter commented lines out of .env in envup

bradurani:
This allows one to comment out lines in a `.env` file using `#` and not have them exported when running `envup`

Without this if your `.env` file is something like

```
SYNC_URL=localhost:4000
# SYNC_URL=production.sync.com
```
and you run `envup`, the 2nd assignment will be exported!

### DIFF
--- a/zsh/functions/envup
+++ b/zsh/functions/envup
@@ -2,7 +2,7 @@
 
 function envup() {
   if [ -f .env ]; then
-    export $(cat .env)
+    export $(sed '/^ *#/ d' .env)
   else
     echo 'No .env file found' 1>&2
     return 1


geoffharcourt:
Nice catch!

cc: @thoughtbot/shell

zsh/functions/envup
```diff
@@ -2,7 +2,7 @@
 
 function envup() {
   if [ -f .env ]; then
-    export $(cat .env)
+    export $(sed '/^#/ d' .env)
```
christoomey: `sed` certainly works here, but it feels a bit odd to me. My lean would be to use `grep` as I associate that with finding / filtering lines, and `sed` with transforming, but that is sort of an arbitrary and perhaps artificial distinction.

If you did want to use `grep`, this could be:

```
export $(grep -v '^#' .env)
```

mike-burns: I'm cool with using sed like that (I guess because I write lines like that in vi?). But how about a more permissive regexp? Is it unreasonable to have e.g. `/^ *#/` -- allow whitespace before the comment?

bradurani:
Regex updated to filter out comments with leading whitespace 😃

mike-burns:
LGTM.

geoffharcourt:
Thanks for contributing, @bradurani!

PR #550, opened 2017-09-28
Update references from ~/.<UTIL>.local to ~/dotfiles-local/<UTIL>.local

tkbrigham:
Based on [this commit](https://github.com/thoughtbot/dotfiles/commit/09d54d448c0d70f2e404e7610cac1d5314f76593?diff=split), it looks like local conf files have moved from (e.g.) `~/.zshrc.local` => `~/dotfiles-local/zshrc.local`.

Unless these references were left in for backwards compatibility, they should reference the new structure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Your `~/dotfiles-local/vimrc.local` might look like this:
 
 If you don't wish to install a vim plugin from the default set of vim plugins in
 `.vimrc.bundles`, you can ignore the plugin by calling it out with `UnPlug` in
-your `~/.vimrc.bundles.local`.
+your `~/dotfiles-local/vimrc.bundles.local`.
 
     " Don't install vim-scripts/tComment
     UnPlug 'tComment'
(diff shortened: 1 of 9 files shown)


geoffharcourt:
Hi @tkbrigham! 

I don't think this is true. I have my personal files in `~/dotfiles-local` like (`~/.dotfiles-local/zshrc.local`), but I then use https://github.com/thoughtbot/rcm to symlink them into place alongside the shared base dotfiles.

iamarmanjon:
Hi @tkbrigham if you look at this commit https://github.com/thoughtbot/dotfiles/commit/8c77a89bc3797ace04fcaab807ce5f6bbccda914#diff-30467b1dec89ba4714905eed4c11b38a. They switch `~/dotfiles-local` to load the overwrite settings first instead of the ones from the repo.

geoffharcourt:
@iamarmanjon my first dotfiles commit! That's a deep track.

I think this is just a misunderstanding, @tkbrigham let me know if you're still having issues.

tkbrigham:
Absolutely right - thanks @geoffharcourt and @iamarmanjon !

PR #551, opened 2017-09-29
Add instruction to overwrite default configurations via dotfiles-local

iamarmanjon:
It's not obvious currently that you can overwrite anything that is being symlinked in `dotfiles` folder with the files you have in `dotfiles-local` as long as you name it the same. So in my case, I wanted to maintain my own vim configuration but I don't want to lose the other configuration that is already set  in place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,9 +99,7 @@ Your `~/dotfiles-local/vimrc.local` might look like this:
     highlight NonText guibg=#060606
     highlight Folded  guibg=#0A0A0A guifg=#9090D0
 
-If you want to use your own `.vimrc` or `.vimrc.bundles` file. You can create your own `vimrc` or `vimrc.bundles` in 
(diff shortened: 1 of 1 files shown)


geoffharcourt:
Being able to completely overwrite files is something I think we would benefit from pointing out, but I think this addition would lead someone to believe that if they wanted to have customizations for Vim that they'd need to abandon the base `vimrc` and `vimrc.bundles` rather than use `vimrc.local` and `vimrc.bundles.local`. 

I think this is on the right track but maybe we could rephrase to say something along the lines of, "if you want your customizations to overwrite rather than add to the files in this repository, you can give them the name of the original file, and rcm will use your file in `dotfiles-local` over the version here." or something like that?

iamarmanjon:
Yes, that is exactly what I meant @geoffharcourt :)

PR #552, opened 2017-10-02
Fixup g function to start with function keywork (like our other functions)

jeckhart:
I was getting the following error when manually sourcing ~/.zshrc:

    /Users/jeckhart/.zsh/functions/g:3: defining function based on alias `g'
    /Users/jeckhart/.zsh/functions/g:3: parse error near `()'

This cleans up the g function definition.

### DIFF
--- a/zsh/functions/g
+++ b/zsh/functions/g
@@ -1,6 +1,6 @@
 # No arguments: `git status`
 # With arguments: acts like `git`
-g() {
+function g() {
   if [[ $# -gt 0 ]]; then
     git "$@"
   else


geoffharcourt:
cc: @thoughtbot/shell

mike-burns:
Why is this a function instead of a command?

jeckhart:
Alias expansion happens before function defs in zsh and can step on the function def. For example:

    alias ll='/bin/ls -F'
    ll { /bin/ls -la "$@" | more }

ll in the function definition is in command position and is expanded as an alias, defining /bin/ls and -F as functions which call /bin/ls which is not the intention.

One workaround is to force the function name by using the function keyword:

    alias ll='/bin/ls -F'
    function ll { /bin/ls -la "$@" | more }

Examples from http://zsh.sourceforge.net/FAQ/zshfaq02.html#23

mike-burns:
Ah sorry, I was too quick with my language.

Can we move this function into the `bin/` dir and make it into a proper command? Does it need to be a function?

geoffharcourt:
@jeckhart bump, can you respond to @mike-burns and we can make a merge or amend decision?

geoffharcourt:
Closing this out for now. @jeckhart if you are still interested in pursuing this, we'll open it back up.

PR #554, opened 2017-10-05
Exclude README in Spanish with RCRC variable

nhutdm:


geoffharcourt:
Hi @nhutdm this is a great idea. Since we might add further translations in the future, could we do this with a pattern? I think `EXCLUDES="README*.md LICENSE"` works.

Here's the relevant rcm doc:
http://thoughtbot.github.io/rcm/lsrc.1.html#EXCLUDE_PATTERN

nhutdm:
Thank @geoffharcourt, it works with exclude pattern. I committed my code.

geoffharcourt:
Verified and merged. Thanks @nhutdm!

PR #555, opened 2017-10-06
Remove Vundle shim and deprecation warning

geoffharcourt:
We've been using vim-plug over Vundle since 5724d12 (April 2015). Users
have had long enough to migrate their `vimrc.bundles.local` to use
vim-plug's `Plug` command.


mike-burns:
NOooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo ok fine.

PR #557, opened 2017-10-20
Exclude ASDF project version file from Git

geoffharcourt:
ASDF respects tool-specific version files through the
`legacy_versions` configuration parameter. This change prevents a
developer's settings from creeping into the repository.

### DIFF
--- a/git_template/info/exclude
+++ b/git_template/info/exclude
@@ -4,3 +4,5 @@
 # exclude patterns (uncomment them if you want to use them):
 # *.[oa]
 # *~
+
+.tool-versions


derekprior:
I don’t think we want to do this. Consider, for example, that Rails projects also now allow the use of node modules via Yarn for assets. `.tool-versions’ allows us to lock to a node version if desired. 

Also consider Phoenix projects, where ASDF is more entrenched as the version manager of choice. `.tool_versions` will often contain the elixir, erlang, and node versions necessary. 

I would rather see us add support for parsing `.tool-versions` to other version managers or something like that. Or maybe we just live with the duplicate ruby version information in yet another location (it’s already duplicated in the Gemfile of our Rails projects).

geoffharcourt:
@derekprior thanks for the feedback, I have some concerns that this wasn't the right approach.

PR #558, opened 2017-10-21
added new shortcuts to gitconfig

chkettu:


geoffharcourt:
Hi @chkettu I think these are fairly personal customizations, so these might be best used as part of your `gitconfig.local`.

PR #559, opened 2017-10-25
Replace CtrlP with FZF

rickpeyton:
This is probably a long shot, but I thought I'd throw it out there and get feedback. If you all pick it up then I won't have to carry a fork 👍

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -79,17 +79,14 @@ if executable('ag')
   " Use Ag over Grep
   set grepprg=ag\ --nogroup\ --nocolor
 
-  " Use ag in CtrlP for listing files. Lightning fast and respects .gitignore
-  let g:ctrlp_user_command = 'ag -Q -l --nocolor --hidden -g "" %s'
-
-  " ag is fast enough that CtrlP doesn't need to cache
-  let g:ctrlp_use_caching = 0
-
   if !exists(":Ag")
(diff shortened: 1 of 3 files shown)


geoffharcourt:
I do this myself, but I've been reluctant to force an extra dependency installation on users. Let's think this over before we make a decision.

christoomey:
I'm also using fzf.vim in place of CtrlP, but similar to Geoff I think this may be a bit much to change in the dotfiles as there are some features that fzf does not support, and some non trivial differences in the interface. Specific features I can think of are fzf not filtering out the current buffer from the file list, and fzf not remembering the previous entry.

rickpeyton:
For anyone that happens upon this PR and wants to give FZF a try... I like the way you guys managed it.

https://github.com/geoffharcourt/dotfiles-local/blob/master/vimrc.bundles.local#L28
https://github.com/geoffharcourt/dotfiles-local/blob/373b3301a48a3417e3e2be9bf611936d1056b51c/vim/plugin/mappings.vim#L48

https://github.com/christoomey/dotfiles/blob/master/vim/rcplugins/fzf#L25-L31

geoffharcourt:
I'm reluctant to force another dependency on users, so I'm going to close this out. FZF will continue on in our local dotfile customizations...

PR #560, opened 2017-10-27
Add ALE fix for Rails ERB templates

### LINKED ISSUE #556: w0rp ale issues with erb files
Has anyone experienced any issues with syntax on .erb files using `w0rp/ale`? I started experiencing this issue today upon updating my `dotfiles` and see that there is already an issue open in their repo. https://github.com/w0rp/ale/issues/580

Here is a screenshot that shows a blank .erb file open with the error. Wondering if anyone has come across this issue and has a fix.

screenshot 2017-10-19 15 53 28

geoffharcourt:
See this discussion on the related ALE issue:
https://github.com/w0rp/ale/issues/580#issuecomment-337676607

This change sets up ALE to use a custom linter for Rails ERB templates to
accommodate the syntax. Without the change, the `erb` linter raises
errors on templates within a Rails application (including blank files).

We define the `g:ale_linters` dictionary conditionally in case the user
has already defined it elsewhere. Users with existing ALE linter
customizations may need to change their definition so that they don't
wipe this setting out.

Close #556.

### DIFF
--- a/vim/plugin/ale.vim
+++ b/vim/plugin/ale.vim
@@ -1,0 +1,13 @@
+call ale#linter#Define('eruby', {
+  \   'name': 'erubylint',
+  \   'executable': 'erb',
+  \   'output_stream': 'stderr',
+  \   'command': "ruby -rerb -e \"puts ERB.new(File.read(%t, encoding: 'BINARY').gsub('<%=','<%'), nil, '-').src\" | ruby -c",
+  \   'callback': 'ale#handlers#ruby#HandleSyntaxErrors',
+  \})
+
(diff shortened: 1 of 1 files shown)


derekprior:
I tried this briefly. It cleaned up the errors I was getting in ERB files while also leaving my Ruby files as error-filled as before. However, I wasn't able to get ERB files to give me *any* errors. Is it supposed to have the same rules as RuboCop (e.g., single quotes, 80 characters, etc)?

geoffharcourt:
It looks like ALE fixed this here: https://github.com/w0rp/ale/pull/1114. Closing.

PR #561, opened 2017-10-27
Include JavaScript in tags

geoffharcourt:
We're doing more projects that feature JavaScript more heavily. This
change brings in some JavaScript setup for Ctags so we can use tags with
JS projects.

Credit to Ray Grasso (@grassdog) for sharing his nice ctags config.
https://github.com/grassdog/dotfiles/blob/6bd36bcb59b57eac28d618f76f21e83d4fc487a8/ctags#L46-L67

Added an exclusion filter for `tmp`.


geoffharcourt:
I've seen a couple warnings about unrecognized regexes in ctags, so I'm going to sort that out before merging this. Thank you @mwenger1 and @christoomey for the reviews.

thorncp:
Is this a duplicate of https://github.com/thoughtbot/dotfiles/pull/486 ?

geoffharcourt:
Closing this for now, we'll revisit if we get a better solution without warnings.

PR #564, opened 2017-11-05
Use long options for when calling `ag` for CtrlP plugin

andyw8:
I propose to change this because:

* It is more self-documenting to use the long options.
* It uses a consistent style, rather than mixing short and long styles.


geoffharcourt:
Sweet! Nice change, I like doing this and making ourselves clearer. Thanks @andyw8.

PR #565, opened 2017-11-07
Change zsh ctrl-n keybinding to be opposite of ctrl-p

turino:
Everywhere where ctrl-p is "previous" I expect ctrl-n to be "next".

Not sure what to do about the `insert-last-word` keybinding -- OK to remove? Map it to something else?

### DIFF
--- a/zsh/configs/keybindings.zsh
+++ b/zsh/configs/keybindings.zsh
@@ -11,7 +11,8 @@ bindkey "^E" end-of-line
 bindkey "^K" kill-line
 bindkey "^R" history-incremental-search-backward
 bindkey "^P" history-search-backward
+bindkey "^N" history-search-forward
 bindkey "^Y" accept-and-hold
-bindkey "^N" insert-last-word
+bindkey "^[." insert-last-word
 bindkey "^Q" push-line-or-edit
 bindkey -s "^T" "^[Isudo ^[A" # "t" for "toughguy"


geoffharcourt:
I've never used this mapping before, so I'm OK with moving it absent a compelling reason to keep what we had.

mike-burns:
Oh, I see: ^N because "next" starts with an n. Got it.

@geoffharcourt if you haven't used `insert-last-word`, you're missing out. Next time you type a command and notice that you're passing the last argument of the prior command, take note. It happens far more often than it should.

The default is down arrow for search forward, up arrow for search backward; there's also ^S for search forward, popular among emacs users.

Anyway, I'm against this change because ^N is in my muscle memory now and I like pressing ^N on the coworkers' keyboards and discovering that it works. As a compromise, I'd be OK with remapping it to another key. I don't remember where I got ^N from.

turino:
@mike-burns yeah "n" is for next and "p" is for previous. I've yet to find a place where ctrl-p works and ctrl-n doesn't do what I expect. It's the default in bash (https://ss64.com/bash/syntax-keyboard.html). It even works in the Chrome address bar (on Mac, anyway).

What does "s" stand for?

mike-burns:
S is for "search", I suppose -- though I don't feel comfortable making claims about key bindings mapping to English words.

turino:
If you do a text search at https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts for "ctrl+p", it always has "ctrl+n" as its opposite. (3 occurrences, 2 of which are for the same command on different systems)

mike-burns:
How about remapping `insert-last-word` to something else? Preferrably a control-something so I don't have to reach too far.

turino:
@mike-burns I set `insert-last-word` to ctrl-b -- does that work for you?

geoffharcourt:
@turino we probably shouldn't use ctrl-b, it's assigned as the tmux prefix: 
https://github.com/thoughtbot/dotfiles/blob/b9b9e7db8c2d9ce57a982aa896e3374cffffb14a/tmux.conf#L34

turino:
could we use ctrl-o?

bernerdschaefer:
The default for insert-last-word in emacs mode for bash/zsh/inputrc is `ESC/meta/alt/option .`:

```
"^[." insert-last-word
```

That's the keybinding I always expect to be available for this operation -- what about bringing that over into the vi-mode?

turino:
OK, I changed the binding to "^[."

@mike-burns is this acceptable?

mike-burns:
I think ^[ conflicts with one of the ways to press ESC ... but I don't use that way so I'm cool with it.

gylaz:
Yep `^[` is an Esc binding in Vim. Some of us use it because of the Mac Touchbar making `Esc` practically unusable.

gylaz:
Yep `^[` is an Esc binding in Vim. Some of us use it because of the Mac Touchbar making `Esc` practically unusable.

gylaz:
Yep `^[` is an Esc binding in Vim. Some of us use it because of the Mac Touchbar making `Esc` practically unusable.

bernerdschaefer:
FWIW, `^[` is literally ESC, not `Ctrl-[`, so you would use `ESC+.` to insert the last word.

Assuming you have "Use option as meta key" set (default in iTerm, under settings in Terminal.app), you can also use `alt+.` to insert last word.

turino:
so this would take away the ability to use `ESC+.` to repeat the last action, which I doubt anyone ever does, but it would give them very confusing behavior if anyone ever tried it. (Which is where I was with `ctrl-n` before this PR).

So... should I just change my local config and close this PR?

geoffharcourt:
This yak is _smoooooth_.

bernerdschaefer:
Ah, I see the problem now for the binding in vi-mode. Fun stuff 😄

turino:
OK, I'm going to close this since no one else seems to use `ctrl-n` the way I do

PR #573, opened 2017-11-21
Use -g instead of --filename-pattern as Ag argument

karunsiri:
While I love how explicit `--filename-pattern` is, the option is unavailable in Linux build, or at least Ubuntu build. 
When running CtrlP on Linux, it will show ag help instead of file list in a directory
ctrlp2

This is because the option is available in Mac OS build. I checked by issuing `ag` command on both of my laptops and found out that it Ag did not alias `--filename-pattern` on Ubuntu machine.

This change should make Ag and CtrlP work together more universally.

_Fixed_
ctrlp fixed

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -80,7 +80,7 @@ if executable('ag')
   set grepprg=ag\ --nogroup\ --nocolor
 
   " Use ag in CtrlP for listing files. Lightning fast and respects .gitignore
-  let g:ctrlp_user_command = 'ag --literal --files-with-matches --nocolor --hidden --filename-pattern "" %s'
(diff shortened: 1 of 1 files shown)


geoffharcourt:
cc: @andyw8 

Should we revert #564?

karunsiri:
@geoffharcourt We may revert that change, yes, but I still like the clarity the long option offers, so I propose we change just one option short. It is inconsistent though. What do you think?

geoffharcourt:
@karunsiri I merged as cad8f46 and updated the commit message to explain why you changed the option. Thank you for your contribution!

andyw8:
@geoffharcourt Sorry, I only just saw this.

It seems odd that the flags differ on Ubuntu. Which version of ag are you running @karunsiri ?

PR #578, opened 2017-12-28
Use consistent syntax for tmux mappings

andyw8:
`bind` is effectively an alias for `bind-key`:

See https://superuser.com/questions/581939/on-tmux-what-is-the-difference-between-bind-and-bind-key

(The file current contains an equal count of `bind` vs `bind-key`, so I could have changed the others to `bind`, but I think it's better to use the full command name).


geoffharcourt:
Thanks @andyw8!

PR #579, opened 2018-01-02
git: Use `zebra` for moved line colors

geoffharcourt:
This change uses a git 2.15 feature to change the diff colors for lines
that have been moved rather than added or deleted. It provides a cue to
readers that a line's content has not changed.

Adding this setting in gitconfig should not break earlier git. (Tested
against 2.14 and the change had no effect.)

https://twitter.com/SmileyKeith/status/946789768506912773

### DIFF
--- a/gitconfig
+++ b/gitconfig
@@ -25,3 +25,5 @@
   autosquash = true
 [include]
   path = ~/.gitconfig.local
+[diff]
+  colorMoved = zebra


keith:
Super glad to see this added here. But would you mind removing my handle from the commit message? It ends up mentioning me a lot as users grab this commit for their dotfiles.

geoffharcourt:
Sure thing, very sorry about that, only meant to assign credit.

keith:
No problem! Just glad more people will use this awesome feature!

gitconfig
```diff
@@ -25,3 +25,5 @@
   autosquash = true
 [include]
   path = ~/.gitconfig.local
+[diff]
+	colorMoved = zebra
```
nickcharlton: Should we match the indentation here?

geoffharcourt: Yeah, that was added when I copy-pasted from the git auto-addition. Nice catch!

geoffharcourt: Hrm, this was weird, it didn't look over-indented when I checked back. Replaced the spaces manually in case there were invisible characters there.

PR #582, opened 2018-02-03
Add co-author prompt to gitmessage template

tysongach:
GitHub now supports trailers for co-authors. The syntax isn't the
easiest to remember, so having it as a prompt when commiting makes the
process easier.

More info: https://github.com/blog/2496-commit-together-with-co-authors


jamescmartinez:
🎊 @tysongach

PR #584, opened 2018-02-16
Restore configurations from old dotfiles

mwenger1:
https://github.com/mwenger1/dotfiles-backup


mwenger1:
This PR can be disregarded. It was intended to be opened against my forked version

PR #585, opened 2018-02-26
Remove vim-bundler to address slow startup

derekprior:
I've experienced slow startup times with vim-bundler off and on for the
past few years. I don't think I use anything that its giving me and in
my pairing I haven't seen anyone else at thoughtbot use `:Bundle`, gem
ctags or anything else that this plugin adds.

See: https://github.com/tpope/vim-bundler/issues/38


geoffharcourt:
I use `:Bopen <gemname>` a lot to dive into the source of a dependency in an app. We could probably alias this to do the same thing without the other initialization burden the plugin brings.

teoljungberg:
I do the same as @geoffharcourt - there has been work done in the plugin to increase its performance. For example: https://github.com/tpope/vim-bundler/commit/04d2980e04c745744f18573213631ab897d6aff5

Issue #594, opened 2018-05-15
mcd causes tab completion conflict with mtools


ArtistsTechGuy:
same problem as https://github.com/robbyrussell/oh-my-zsh/issues/1895

geoffharcourt:
@ArtistsTechGuy this appears to be fixed now by #595? Please reopen if it's still an issue.

PR #595, opened 2018-05-15
Improve successive startup time; fix conflict.

ArtistsTechGuy:
- Don't update completion dump if it has been updated within 24hrs.
- Unload Completion for zsh bundled function mcd to prevent conflict. #594


ArtistsTechGuy:
The other way to fix it would be to rename the function to something like mkcd.  Also, should I split this out into 2 commits then?  I don't have much experience contributing to projects so advice is appreciated.

geoffharcourt:

Issue #588, opened 2018-03-10
.agignore deprecated in favour of .ignore as of ag 2.0.0


andyw8:
https://github.com/ggreer/the_silver_searcher/wiki/Advanced-Usage

(2.0.0 was [released](https://github.com/ggreer/the_silver_searcher/releases) in June 2017).

PR #596, opened 2018-06-07
Move from agignore to ignore

geoffharcourt:
Per @andyw8 in #588:
https://github.com/ggreer/the_silver_searcher/wiki/Advanced-Usage

This convention has been part of Ag for over a year now, so we should be
able to safely move to it. Users who are unable to use a recent version
can continue to use `.agignore` through a local configuration override.


geoffharcourt:
Updated to use `.gitignore` in 85938fe89d2145e791051d912cc1d96d09b1ace9.

PR #602, opened 2018-07-13
Set vim syntax for local dotfiles overrides

stevehanson:
This adds vim syntax highlighting for `.local` dotfile override files. It seems like it makes sense to have these here since the Readme suggests setting up these files.


geoffharcourt:

PR #612, opened 2018-10-24
Make _load_settings() more concise

benjamineskola:
- Use the `~` glob operator to exclude *.zwc insead of a separate test
  (requires `setopt extendedglob`).
- Exclude both `pre` and `post` in one case alternative.


geoffharcourt:
cc: @thoughtbot/shell

PR #613, opened 2018-10-27
Add git alias to prune old branches: git pob

kenhkelly:
`git pob` reaches out to the remote to find branches that have been deleted. If they have been deleted from the remote and you are tracking them, then those local branches will be deleted. This is completely safe on branches that you are tracking that have not been deleted. Local branches that you have not pushed up are also completely safe.

### DIFF
--- a/gitconfig
+++ b/gitconfig
@@ -11,6 +11,7 @@
   ci = commit -v
   co = checkout
   pf = push --force-with-lease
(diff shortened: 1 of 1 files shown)


mikegee:
@kenhkelly could you fill in a bit of a description of this alias? One detail I'm curious about is the risk of accidentally deleting code.

kenhkelly:
@mikegee, @geoffharcourt  Updated the description

PR #614, opened 2018-10-27
Add cd helper aliases to traverse backwards quickly.

kenhkelly:


mikegee:
I think `cd ..` and just `..` already work the same without these. Can you check if that's true for you?

I don't think I setup any particular setting for that, I think `zsh` automatically changes directory if you leave off the `cd`.

kenhkelly:
I think you are right, I will remove and push up

composerinteralia:
Thanks for the PR! These dotfiles are a based on what many of us use here at thoughtbot. Since we aren’t using these `cd` aliases, I am going to close this.

PR #618, opened 2018-11-29
Add UTF-8 encoding to vimrc

carlesjove:
I started getting errors after a recent Vim upgrade because of this line:

`set list listchars=tab:»·,trail:·,nbsp:·`

Adding UTF-8 encoding fixes it.

```text
:h encoding

NOTE: For GTK+ 2 or later, it is highly recommended to set 'encoding'
        to "utf-8".  Although care has been taken to allow different values of
        'encoding', "utf-8" is the natural choice for the environment and
        avoids unnecessary conversion overhead.  "utf-8" has not been made
        the default to prevent different behavior of the GUI and terminal
        versions, and to avoid changing the encoding of newly created files
        without your knowledge (in case 'fileencodings' is empty).
```

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -1,3 +1,5 @@
+set encoding=utf-8
+
 " Leader
 let mapleader = " "
 


vimrc
```diff
@@ -1,3 +1,5 @@
+set encoding=utf-8
+
```
carlesjove: According to documentation, `scriptencoding utf-8` could be added below that too, to specify encoding used in _this_ file. However, without `encoding` weird characters are being shown instead of the middle dot, so I found that this line is enough.

geoffharcourt:
From the documentation you quoted, will this cause problems for terminal-vim users?

cc: @thoughtbot/shell

carlesjove:
I am a terminal-vim user (Vim 8.x, iTerm 2 for more details) and this is
the only thing that solved this issue for me. I had been using these
dotfiles as well as Vim for a long time, and the issue only appeared after
a recent vim upgrade.

jonathanbranam:
Is there a reason this change has not been merged? I just setup a new macOS laptop on Mojave. Everything was fine with the system vim which is 8.0. But when I updated to the latest vim from homebrew 8.1 this error popped up immediately. I can  I use vim in iTerm with tmux primarily, if that matters.

```
VIM - Vi IMproved 8.1 (2018 May 18, compiled Jan 14 2019 22:41:50)
macOS version
Included patches: 1-750
Compiled by Homebrew
```

I tested only with `scriptencoding utf-8` which fixes the error but since files other than `.vimrc` are not opened in utf-8 the middle dot does not render in any other file.

PR #622, opened 2018-12-30
Clarify ctags mapping

andyw8:
`vim/plugin/ctags.vim` shells out to ctags, but this could be Universal
Ctags (see #605).


geoffharcourt:
:100:

PR #626, opened 2019-02-28
has-session - only exact matches

rofrol:
from man
>If the session name is prefixed with an ‘=’, only an exact match is accepted (so ‘=mysess’ will only match exactly ‘mysess’, not ‘mysession’).


geoffharcourt:

PR #629, opened 2019-04-23
Account for `asdf` being installed by Homebrew

danbee:
If `asdf` has been installed via Homebrew the completions file is in a
different place. This PR takes that possibility into account.

### DIFF
--- a/zsh/configs/post/path.zsh
+++ b/zsh/configs/post/path.zsh
@@ -1,12 +1,16 @@
 # ensure dotfiles bin directory is loaded first
 PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
-# load ASDF, either from home dir or brew install
+# Try loading ASDF from the regular home dir location
 if [ -f "$HOME/.asdf/asdf.sh" ]; then
   . "$HOME/.asdf/asdf.sh"
+# It's not in the home dir, let's try the default Homebrew location
 elif [ -f "/usr/local/opt/asdf/asdf.sh" ]; then
   . "/usr/local/opt/asdf/asdf.sh"
-elif which brew >/dev/null && ASDF_PREFIX=$(brew --prefix asdf); then
+# Not there either! Last try, let's see if Homebrew can tell us where it is
+elif which brew >/dev/null &&
+  ASDF_PREFIX=$(brew --prefix asdf) &&
+  [ -f "$ASDF_PREFIX/asdf.sh" ]; then
   . "$ASDF_PREFIX/asdf.sh"
 fi
 
--- a/zsh/configs/post/path.zsh
+++ b/zsh/configs/post/path.zsh
@@ -1,9 +1,11 @@
 # ensure dotfiles bin directory is loaded first
 PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
-# load ASDF, falling back to rbenv if not available
-if [ -d "$HOME/.asdf" ]; then
+# load ASDF, either from home dir or brew install
+if [ -f "$HOME/.asdf/asdf.sh" ]; then
   . $HOME/.asdf/asdf.sh
+elif ASDF_PREFIX=$(brew --prefix asdf); then
+  . $ASDF_PREFIX/asdf.sh
 fi
 
 # mkdir .git/safe in the root of repositories you trust
--- a/zsh/configs/post/path.zsh
+++ b/zsh/configs/post/path.zsh
@@ -3,9 +3,11 @@ PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
 # load ASDF, either from home dir or brew install
 if [ -f "$HOME/.asdf/asdf.sh" ]; then
-  . $HOME/.asdf/asdf.sh
-elif ASDF_PREFIX=$(brew --prefix asdf); then
-  . $ASDF_PREFIX/asdf.sh
+  . "$HOME/.asdf/asdf.sh"
+elif [ -f "/usr/local/opt/asdf/asdf.sh" ]; then
+  . "/usr/local/opt/asdf/asdf.sh"
+elif which brew >/dev/null && ASDF_PREFIX=$(brew --prefix asdf); then
+  . "$ASDF_PREFIX/asdf.sh"
 fi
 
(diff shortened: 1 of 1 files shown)


zsh/configs/post/path.zsh
```diff
@@ -1,9 +1,11 @@
 # ensure dotfiles bin directory is loaded first
 PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
-# load ASDF, falling back to rbenv if not available
-if [ -d "$HOME/.asdf" ]; then
+# load ASDF, either from home dir or brew install
+if [ -f "$HOME/.asdf/asdf.sh" ]; then
   . $HOME/.asdf/asdf.sh
+elif ASDF_PREFIX=$(brew --prefix asdf); then
```
eebs: I'm not great with shell scripting, but I presume this checks if the directory exists? When I run `brew --prefix asdf` on my machine where asdf was not installed via brew, it still returns `/usr/local/Cellar/asdf/0.7.1` even though that file is not there. Does the `elif` return false if the variable assignment fails? I did notice that running `$(brew --prefix asdf)` does give an error (and a 127 exit), so I presume that is falsey.

mike-burns:
```
$ if y=$(brew --prefix asdf); then
> echo hi
> fi
sh: 3: brew: not found
```

I got it to be quieter with a `2>/dev/null` inside the parens.

Is it possible that `brew --prefix asdf` can produce multiple lines of output? I don't actually know what a brew is.

Also also, what do we do about the embedded newline?

All of this line relies on the return value of `brew --prefix asdf`. Is it possible to use `test`(1) here instead and set the variable outside of this line?

danbee: Oh, interesting. I assumed this command would not return anything unless `asdf` was actually installed. I'm gonna have to rethink this.

danbee: Thanks Mike. I did not consider the possibility that `brew` might not be installed! (It's a package manager for the Mac, similar to `apt`). I need to rework this anyway because `brew --prefix asdf` is not a reliable way to test what I'm trying to test for. I'll look up documentation for `test`.

mike-burns: Another spelling of `test` is `[` -- that's how we normally spell it in these files, but the manpage will be under `test`.

rjdellecese: Hey @danbee, thanks for opening this PR! I just got a new machine and installed `asdf` with Homebrew, so I've been running into this issue myself.

What do you think about the following to address the concerns raised by Mike and Eebs?

```bash
# load ASDF, falling back to rbenv if not available
if [ -f "$HOME/.asdf/asdf.sh" ]; then
  . $HOME/.asdf/asdf.sh
elif command -v brew > /dev/null; then
  ASDF_EXECUTABLE="$(brew --prefix asdf)/asdf.sh"

  if [ -f "$ASDF_EXECUTABLE" ]; then
    . "$ASDF_EXECUTABLE"
  fi
fi
```

rjdellecese: Or, much more simply

```zsh
# load ASDF, falling back to rbenv if not available
if [ -f "$HOME/.asdf/asdf.sh" ]; then
  . $HOME/.asdf/asdf.sh
elif [ -f "/usr/local/opts/asdf/asdf.sh" ]; then
  . "/usr/local/opts/asdf/asdf.sh"
fi
```

since `brew --prefix` is agonizingly slow and for the vast majority of Homebrew installs the prefix will be `/usr/local`.

zsh/configs/post/path.zsh
```diff
@@ -1,9 +1,17 @@
...
-  . $HOME/.asdf/asdf.sh
+# Try loading ASDF from the regular home dir location
+if [ -f "$HOME/.asdf/asdf.sh" ]; then
+  . "$HOME/.asdf/asdf.sh"
+# It's not in the home dir, let's try the default Homebrew location
+elif [ -f "/usr/local/opt/asdf/asdf.sh" ]; then
+  . "/usr/local/opt/asdf/asdf.sh"
+# Not there either! Last try, let's see if Homebrew can tell us where it is
+elif which brew >/dev/null &&
+  ASDF_PREFIX=$(brew --prefix asdf) &&
```
danbee: This works but `brew` takes about a second to return here. I'd love to figure out a faster way to check whether a package is installed.

mike-burns: That's pretty slow. Is that `brew` startup time? Or is it searching something? Is `brew ls --versions asdf` faster?

Is it looping through all your installed packages? https://github.com/Homebrew/brew/blob/14755541b0e786e190d415a6d9d74c2e5023802a/Library/Homebrew/extend/ARGV.rb#L29-L35

What if we made a claim like:

> If you changed your Homebrew installation path away from the default, you now have to deal with the consequences: a slower shell startup time. Set `ASDF_PREFIX` to improve this again.

And then start the entire conditional off with trying to source `"$ASDF_PREFIX/asdf.sh"`.

?

danbee: Sadly all `brew` commands take around a second to come back, and it doesn't set any environment variables that we can check either.

I wonder if it might be better to leave this part out entirely? Then if somebody installs Homebrew somewhere else and installs `asdf` it's up to them to source the completions file in their local config file.

zsh/configs/post/path.zsh
```diff
@@ -1,9 +1,13 @@
...
 PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 
-# load ASDF, falling back to rbenv if not available
-if [ -d "$HOME/.asdf" ]; then
-  . $HOME/.asdf/asdf.sh
+# Try loading ASDF from the regular home dir location
+if [ -f "$HOME/.asdf/asdf.sh" ]; then
+  . "$HOME/.asdf/asdf.sh"
+elif which brew >/dev/null &&
+  BREW_DIR=$(which brew | sed 's/\/bin\/brew//') &&
```
mike-burns: Maybe `dirname` instead.

```suggestion
  BREW_DIR="$(dirname $(which brew)/.." &&
```

PR #632, opened 2019-05-14
Add easier tmux binding keys - to split window horizontally and vertically

ysfdev:
PR adds adds custom tmux binding keys to split window horizontally and vertically. 

Tmux Invocation for horizontal split: 
```
Ctrl + b + - 
```

Tmux Invocation for vertical split: 
```
Ctrl + b + /
```

### DIFF
--- a/tmux.conf
+++ b/tmux.conf
@@ -30,6 +30,9 @@ set -g status-right ''
 # allow ctrl + b + - to split window horizontally
 bind - split-window -v 
 
+# allow ctrl + b + | to split window vertically
+bind split-window -h
+
 # increase scrollback lines
 set -g history-limit 10000
 
--- a/tmux.conf
+++ b/tmux.conf
@@ -30,7 +30,8 @@ set -g status-right ''
 # prefix + - split window horizontally
 bind-key - split-window -v  -c '#{pane_current_path}'
 
-# prefix + \\ split window vertically
+# prefix + \ split window vertically
+# Note: As of tmux 3.0, we need to escape backslashes
 bind-key \\ split-window -h  -c '#{pane_current_path}'
 
 # increase scrollback lines
--- a/tmux.conf
+++ b/tmux.conf
@@ -31,7 +31,7 @@ set -g status-right ''
 bind - split-window -v 
 
 # allow ctrl + b + | to split window vertically
-bind split-window -h
+bind | split-window -h
 
 # increase scrollback lines
 set -g history-limit 10000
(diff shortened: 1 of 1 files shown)


geoffharcourt:
I use these in my own mappings, but my only hesitation is over making this choice for others.

rspeicher:
@geoffharcourt Is there something bound to these invocations by default? If not I'd consider it a pretty safe choice, while being much more intuitive than the `"` and `%` defaults.

ysfdev:
Hi @geoffharcourt, 

Thanks for taking a look. Based on what @rspeicher has stated above, do you have any other concerns?

tmux.conf
```diff
@@ -27,6 +27,12 @@ set -g status-fg '#aaaaaa'
 set -g status-left ''
 set -g status-right ''
 
+# allow ctrl + b + - to split window horizontally
+bind - split-window -v
```
geoffharcourt: A small change here will open the new pane at the same path as the current pane:

```
bind-key - split-window -v  -c '#{pane_current_path}'
```

tmux.conf
```diff
@@ -27,6 +27,12 @@ set -g status-fg '#aaaaaa'
 set -g status-left ''
 set -g status-right ''
 
+# allow ctrl + b + - to split window horizontally
+bind - split-window -v 
+
+# allow ctrl + b + | to split window vertically
+bind | split-window -h
```
geoffharcourt: Not sure if it matters, but I did this on `\` instead of `|` to avoid an extra keypress on a very common operation. tmux 3.x has a breaking change where backslashes need to be escaped in the config file:

```
bind-key \\ split-window -h  -c '#{pane_current_path}'
```

tmux.conf
ysfdev: Makes sense. I updated the bindings. Thanks

tmux.conf
```diff
@@ -27,6 +27,12 @@ set -g status-fg '#aaaaaa'
 set -g status-left ''
 set -g status-right ''
 
+# prefix + - split window horizontally
+bind-key - split-window -v  -c '#{pane_current_path}'
+
+# prefix + \\ split window vertically
```
geoffharcourt: I think your last push might have removed the edit, this comment doesn't match the keymapping as it has the escaping in it.

composerinteralia:
> I use these in my own mappings, but my only hesitation is over making this choice for others.

For what it is worth, I use these mappings and have seen some other thoughtbot folks using them as well. I am a fan of including them in our dotfiles.

tmux.conf
ysfdev: Yes, I have updated now, and also include note about escaping requirement.
Thanks!

PR #633, opened 2019-05-15
Remove meta files deletion instructions from README

vnbrs:
These files are not copied by rcm anymore as they are ignored by the [`rcrc`](https://github.com/thoughtbot/dotfiles/blob/master/rcrc) file.


geoffharcourt:
I think this PR may be in error. These do not appear to be instructions to exclude these files, but rather an explanation of what the included `rcrc` does.

PR #636, opened 2019-07-12
Replace CtrlP with FZF

SViccari:
We noticed this has been attempted once before but decided it was time to re-open the discussion. The previous discussion can be found here: https://github.com/thoughtbot/dotfiles/pull/559.

Summary:
Using CtrlP has proven to be slower and has fewer features. We've been
switching over to FZF on a frequent enough basis that it feels
appropriate to make FZF the new default for fuzzy searching.

By swapping out Ctrlp in favor of FZF, this should not break anyone's
fuzzy search but it will replace the tool that executes the search.

We believe that the dependency management concern is addressed by
first checking if FZF is available as an executable with the fallback of vim
plug's installation handling the installation of FZF. FZF is also written by
the creator and maintainer of Vim-Plug.

Co-authored-by: Chris Toomey <chris@ctoomey.com>

### DIFF
--- a/vimrc.bundles
+++ b/vimrc.bundles
@@ -16,7 +16,12 @@ call plug#begin('~/.vim/bundle')
 
 " Define bundles via Github repos
 Plug 'christoomey/vim-run-interactive'
-Plug 'ctrlpvim/ctrlp.vim'
+if executable('fzf')
+  Plug '/usr/local/opt/fzf'
+else
+  Plug 'junegunn/fzf', { 'dir': '~/.fzf', 'do': './install --all' }
+endif
+Plug 'junegunn/fzf.vim'
 Plug 'elixir-lang/vim-elixir'
 Plug 'fatih/vim-go'
 Plug 'janko-m/vim-test'
--- a/README-ES.md
+++ b/README-ES.md
@@ -182,7 +182,7 @@ sin importar de nombre del archivo:
 
 Configuración [vim](http://www.vim.org/):
 
-* [Ctrl-P](https://github.com/ctrlpvim/ctrlp.vim) para hallazgo difuso de archivos/buffer/tags.
+* [fzf](https://github.com/junegunn/fzf.vim) para hallazgo difuso de archivos/buffer/tags.
 * [Rails.vim](https://github.com/tpope/vim-rails) para una mejor navegación de la estructura
 de archivos de Rails via `gf` y `:A` (alterno), `:Rextract` parciales,`:Rinvert` migraciones, etc.
 * Ejecuta muchos tipos de pruebas [desde vim]([https://github.com/janko-m/vim-test)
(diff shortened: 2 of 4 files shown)


christoomey:
Discussing this with Steph today, I very much share her desire to see this added to the core. We've both spent time on a number of occasions migrating folks from CtrlP -> fzf, and it would be lovely if we didn't have to do that anymore.

I understand and largely agree with the hesitation around the complexity of external dependency, but I'm hopeful that we can find a configuration that is sufficiently self-contained to make this work.

christoomey:
> We might want to make a note that folks shouldn't also install FZF via Homebrew to avoid a double-install

geoffharcourt:
Whoops, see that now. This is great!

SViccari:
@geoffharcourt Thank you for the review! 🐳

vimrc.bundles
```diff
@@ -16,7 +16,12 @@ call plug#begin('~/.vim/bundle')
 
 " Define bundles via Github repos
 Plug 'christoomey/vim-run-interactive'
-Plug 'ctrlpvim/ctrlp.vim'
+if executable('fzf')
+  Plug '/usr/local/opt/fzf'
```
gylaz: For me on Mac `fzf` is in `/usr/local/bin/fzf` (installed via HomeBrew).

gylaz: Ah, never mind. I thought this was supposed to be the path of the executable.

SViccari: @gylaz Hi! Is this working for you? It seems you've resolved your concern but I just want to check-in and make sure that's true. :)

gylaz: 👋Thanks for checking in @SViccari! I got it working according with this PR.

geoffharcourt:
@SViccari and @christoomey I ran into some problems with this after the merge. I thought it was due to my prior setup, but now I don't think so. Here's what happens:

1. I run `PlugInstall`. I have an `fzf` executable available, but it's not from Homebrew (it's from the vim plugin). `PlugInstall` doesn't give any failure message.
2. Most fzf wrapper functions now fail. No combination of `PlugClean`, `PlugInstall`, or removing files or directories will fix this situation.

*The issue is that I have `$HOME/.fzf/bin` in my path.* I don't think this is an uncommon occurrence, because it's the easiest way to leverage fzf that's been installed by/with Vim rather than to have two copies installed (one for Vim and one from Homebrew).

I'm not sure if the solution is to add `$HOME/.fzf/bin` to the `PATH` or not.

eebs:
I ran into the same issue as @geoffharcourt. I'm able to temporarily fix the issue by adding

```
Plug 'junegunn/fzf', { 'dir': '~/.fzf', 'do': './install --all' }
```

to my local overrides.

FWIW having two copies of fzf installed (one in `$HOME` and on via Brew) seemed to work just fine. We might be able to take an opinionated stance and simply remove the conditional and always install fzf to `~/.fzf`

PR #642, opened 2019-09-26
Aligning source loading with documentation

eliza-jane:
Noticed that gitconfig and zshrc were loading from the wrong spots if
one was following the documentation to set up their dotfiles. There might be more like this that I am happy to also fix, but these were the immediate ones that I noticed.

### DIFF
--- a/gitconfig
+++ b/gitconfig
@@ -24,6 +24,6 @@
 [rebase]
   autosquash = true
 [include]
-  path = ~/.gitconfig.local
+  path = ~/dotfiles-local/gitconfig.local
 [diff]
   colorMoved = zebra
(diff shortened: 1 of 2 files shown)


eebs:
After you run `rcup`, the files in your local dotfiles will be symlinked to your home directory, so I think the paths are correct as is. You should end up with something like

```
.zshrc -> /Users/eebs/dotfiles/zshrc
.zshrc.local -> /Users/eebs/dotfiles-local/zshrc.local
```

This may be slightly different if you've installed either the thoughtbot dotfiles or your own local dotfiles to a location other than your home directory.

eliza-jane:
Wow, I feel like a derp. Thanks, @eebs ! I ran `rcup` but I also had to wipe my development environment last week and start over agin, so I imagine the order didn't line up right. :/

PR #643, opened 2019-10-01
Clean up adjacent aliases with backslashes

rickyjon:
Some Bash built-ins like export, alias can be run once instead of multiple times by using multiple arguments instead of calling it 6 times in a row for 6 alias. I also think the config looks nicer as a result.

### DIFF
--- a/aliases
+++ b/aliases
@@ -1,16 +1,16 @@
 # Unix
-alias ll="ls -al"
-alias ln="ln -v"
-alias mkdir="mkdir -p"
-alias e="$EDITOR"
-alias v="$VISUAL"
+alias ll="ls -al" \
+ln="ln -v" \
+mkdir="mkdir -p" \
+e="$EDITOR" \
+v="$VISUAL"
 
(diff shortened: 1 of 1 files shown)


mikegee:
Could you please write a justification in the PR's description field up there ⬆️

aliases
```diff
@@ -1,16 +1,16 @@
...
+v="$VISUAL"
 
 # Bundler
 alias b="bundle"
 
 # Rails
-alias migrate="rake db:migrate db:rollback && rake db:migrate db:test:prepare"
-alias s="rspec"
+alias migrate="rake db:migrate db:rollback && rake db:migrate db:test:prepare" \
+s="rspec"
```
geoffharcourt: I find this harder to follow than before.

composerinteralia:
Thanks for the contribution. I agree that this is a bit more difficult to follow, and it would also make for noisier diffs if we ever change which alias is first or last in each section, so I am going to close this.

PR #647, opened 2019-10-11
Revise how we check for existing FZF executable

SViccari:
Summary:
This change addresses an error that occurs when we've instructed vim to
reference the wrong path for an existing FZF executable.

Additional Details:
In the previous approach, when checking if FZF has already been
installed, we make the assumption that if FZF is `executable` it was
installed via Homebrew. Then, based on that assumption, we set vim's
`runtimepath` to reference the directory where Homebrew creates symlinks
for installed packages (`/user/local/opt`).

Because of this assumption, anyone that has FZF installed in a different
directory, and available as an executable in their $PATH, attempting to
use FZF within vim will result in an error. The reason vim errors is
because `Plug` has instructed vim's `runtimepath` to look in the wrong
location for FZF (in this case, we told vim to look for FZF within
`/user/local/opt` when the FZF executable lives somewhere else).

The revised approach still attempts to find an existing FZF package
before proceeding with installing FZF.

### DIFF
--- a/vimrc.bundles
+++ b/vimrc.bundles
@@ -16,11 +16,15 @@ call plug#begin('~/.vim/bundle')
 
 " Define bundles via Github repos
 Plug 'christoomey/vim-run-interactive'
-if executable('fzf')
+
+" If fzf has already been installed via Homebrew, use the existing fzf
+" Otherwise, install fzf. The `--all` flag makes fzf accessible outside of vim
+if isdirectory("/usr/local/opt/fzf")
   Plug '/usr/local/opt/fzf'
 else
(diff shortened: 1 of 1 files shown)


SViccari:
@geoffharcourt @eebs I believe this will address the errors you encountered with the recent switch from CtrlP to FZF: https://github.com/thoughtbot/dotfiles/pull/636#issuecomment-523725629 and https://github.com/thoughtbot/dotfiles/pull/636#issuecomment-523910502. 

I considered taking the opinionated approach of always installing FZF to `~/.fzf` but I still like the idea of attempting to use someone's existing package (installed via Homebrew) before installing a new package.

SViccari:
Thank you @eebs, @geoffharcourt and @christoomey for reviewing 🐳

Issue #634, opened 2019-05-21
`_load_settings` needs `set


serixscorpio:
When bringing my dotfiles up-to-date including the [Make _load_settings() more concise]( https://github.com/thoughtbot/dotfiles/commit/46af61168b07f3d738d461eb638eb5824bbeb813) commit, I noticed my prompt turned into 
![2019-05-21_10-45](https://user-images.githubusercontent.com/1740566/58106167-a93aa280-7bb5-11e9-95d6-81e7a4cafed1.png)
It turned out I had to `setopt extendedglob` as noted by the commit message.  I ended up adding this configuration in my `zshenv.local` so by the time `zshrc` runs `_load_settings`, the glob can work correctly.  Just wondering if it would make sense for `setopt extendedglob` to be in thoughtbot dotfiles (zshenv or zshrc?) instead of requiring custom config for `_load_settings` to work?

geoffharcourt:
cc: @thoughtbot/shell

PR #648, opened 2019-10-18
Move extendedglob from Options.zsh to zshrc

rcedwards:
### Issue
This addresses the issue pointed out in https://github.com/thoughtbot/dotfiles/issues/634

### Description
Introduced in https://github.com/thoughtbot/dotfiles/commit/46af61168b07f3d738d461eb638eb5824bbeb813, the `_loadSettings()` func requires `extendedglob`. 

As such, its been moved from `options.zsh` to the `zshrc` file.

### DIFF
--- a/zsh/configs/options.zsh
+++ b/zsh/configs/options.zsh
@@ -2,8 +2,5 @@
 setopt autocd autopushd pushdminus pushdsilent pushdtohome cdablevars
 DIRSTACKSIZE=5
 
-# Enable extended globbing
-setopt extendedglob
-
 # Allow [ or ] whereever you want
 unsetopt nomatch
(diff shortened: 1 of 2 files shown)


eebs:
This option is currently set in `zsh/configs/options.zsh`

https://github.com/thoughtbot/dotfiles/blob/d27cceb0be780a0fc84790ba833a9dd9faa2b62a/zsh/configs/options.zsh#L6

It used to be set before `_load_settings`, but was moved in https://github.com/thoughtbot/dotfiles/pull/430

Perhaps we need to move it before `_load_settings` as you have done, and remove it from `options.zsh`?

rcedwards:
Seems appropriate given its an option required by the function within `zshrc` itself and runs well before the loading of `options.zsh`.

I'll update my PR to match.

opdavies:
This seems to fix my issue in https://github.com/opdavies/dotfiles/issues/13.

PR #649, opened 2019-10-25
Adds a commit-msg git hook

eebs:
People may wish to configure a `commit-msg` hook in their local
dotfiles.

This allows them to do so by adding a [`commit-msg` git hook][1] that
checks for the presence of a local `commit-msg` hook, and runs it if
present. This hook follows the existing pattern we use for other git
hooks.

[1]: https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks

### DIFF
--- a/git_template/hooks/commit-msg
+++ b/git_template/hooks/commit-msg
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+local_hook="$HOME"/.git_template.local/hooks/commit-msg
+
+if [ -f "$local_hook" ]; then
+  . "$local_hook"
+fi


git_template/hooks/commit-msg
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+local_hook="$HOME"/.git_template.local/hooks/commit-msg
+
+if [ -f "$local_hook" ]; then
+  . "$local_hook"
+fi
+
```
geoffharcourt: Let's remove this whitespace here.

eebs: Good catch. Fixed in https://github.com/thoughtbot/dotfiles/pull/649/commits/eeb52a87f6d869d8790936f8dd628c08744d0b5d

PR #652, opened 2019-11-04
Fix gitconfig.local path

cawcaw:
Because of readme:
https://github.com/thoughtbot/dotfiles/blob/master/README.md#L71

### DIFF
--- a/gitconfig
+++ b/gitconfig
@@ -24,6 +24,6 @@
 [rebase]
   autosquash = true
 [include]
-  path = ~/.gitconfig.local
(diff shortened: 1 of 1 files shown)


geoffharcourt:
I don't think this is incorrect. `rcm` should be symlinking that file to the local spot, and it's not a requirement for users to have their customizations sourced from `dotfiles-local`.

cawcaw:
@geoffharcourt yes... now it seems to me that I created this request hastily, sorry 😓

Issue #655, opened 2020-01-09
Error detected while processing ~/.vimrc: line  176: E474: Invalid argument: diffopt+=vertical


jkereako:
Hi there,

I'm running Vim 8.1 on macOS Catalina. When I open Vim, I always get this error message.

> Error detected while processing [~/.vimrc: line  176](https://github.com/thoughtbot/dotfiles/blob/master/vimrc#L176): E474: Invalid argument: diffopt+=vertical Press ENTER or type command to continue

Removing the "+" seemed to have silenced the error, but, because I'm a Vim neophyte I'm not convinced that this is the correct solution.

jimrpy:
The same error.

ashwinvidiyala:
Same here. Did anyone ever find a fix to this?

micahjsmith:
I identified the problem in wrote about it here: https://www.micahsmith.com/blog/2019/11/fixing-vim-invalid-argument-diffopt-iwhite/

In short: stock vim on Catalina does not support `diffopt=internal` even though the default options include `internal`. Thus `:set diffopt+=vertical` is equivalent to `:set diffopt=internal,filler,vertical` which is an error because `internal` is not supported. One solution is to modify your .vimrc as follows:

```vim
if &diff
    set diffopt-=internal
    set diffopt+=vertical
endif
```

ashwinvidiyala:
@micahjsmith Sweet! Thanks :)

pawan2901:
Thanks, it's working for me.

PR #660, opened 2020-03-09
Fix E474: Invalid argument: diffopt+=vertical

brandoncordell:
#655

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -173,7 +173,9 @@ set spellfile=$HOME/.vim-spell-en.utf-8.add
 set complete+=kspell
 
 " Always use vertical diffs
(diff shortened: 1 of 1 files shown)


vimrc
```diff
@@ -173,7 +173,9 @@ set spellfile=$HOME/.vim-spell-en.utf-8.add
 set complete+=kspell
 
 " Always use vertical diffs
-set diffopt+=vertical
+if &diff
+  set diffopt+=vertical
+endif
```
rcreasi, line 178: That doesn't work but [this does](https://github.com/thoughtbot/dotfiles/issues/655#issuecomment-605019271)!

PR #661, opened 2020-04-30
Confirm changes after rebasing

jmeinerz:
Will prompt you with the diff resulting for your rebase and ask whether
you want to keep it or reset the branch to how it was before your rebase
started.

Adapted from:
https://thoughtbot.com/blog/confirm-your-changes-at-the-end-of-a-git-rebase
https://github.com/jmeinerz/rebasing-hook

### DIFF
--- a/git_template/hooks/confirm-changes
+++ b/git_template/hooks/confirm-changes
@@ -10,28 +10,20 @@ if [ -f $rebasing_file ]; then
     git diff $sha_pre_rebase
 
     while true; do
-      echo  "Do you want to keep the changes? [Y/n] "
+      printf  "Do you want to keep the changes? [Y/n] "
       read -r answer < /dev/tty
-      echo ""
 
       case $answer in
-        [Yy])
-          echo "Cool! All changes kept"
-          break
-          ;;
-        [Nn])
+        [Nn]*)
           echo "Discarding changes..."
-          echo ""
           git reset --hard $sha_pre_rebase
           break
           ;;
         *)
-          echo "Answer with Y|y|N|n and press return"
+          break
           ;;
       esac
     done
-  else
-    echo "Your rebase caused no code changes, good for you"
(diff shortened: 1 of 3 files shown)


jmeinerz:
@mike-burns I created this after what we decided in https://github.com/thoughtbot/robots.thoughtbot.com/pull/1754#event-3285741214

I'd also be happy to move the repo from my account to the thoughtbot organisation if that makes sense. What do you think?

git_template/hooks/confirm-changes
```diff
@@ -0,0 +1,38 @@
...
+    git diff $sha_pre_rebase
+
+    while true; do
+      echo  "Do you want to keep the changes? [Y/n] "
+      read -r answer < /dev/tty
+      echo ""
+
+      case $answer in
+        [Yy])
+          echo "Cool! All changes kept"
```
mike-burns, line 19: A bit noisy for doing nothing. I feel like I'd understand that if I interactively answered "y" do "do you want to keep these changes", the result is that my changes would be kept.

git_template/hooks/confirm-changes
```diff
@@ -0,0 +1,38 @@
...
+
+if [ -f $rebasing_file ]; then
+  sha_pre_rebase=$(cat $rebasing_file)
+
+  if [ "$(git diff $sha_pre_rebase)" ]; then
+    echo "Your rebase resulted in code changes"
+    git diff $sha_pre_rebase
+
+    while true; do
+      echo  "Do you want to keep the changes? [Y/n] "
```
mike-burns, line 13: Could use `printf` to keep the cursor on the same line.

```suggestion
      printf  "Do you want to keep the changes? [Y/n] "
```

geoffharcourt:
@mike-burns Maybe folks can try it for a few weeks and see how it feels?

PR #669, opened 2020-06-14
Accept optional branch name (to facilitate moving to "main")

danbee:
There's been a recent movement to rename the "master" branch in git
repositories to "main" to avoid to potentially problematic word
"master". Our `merge-branch` script was hard coded to "master". This PR
makes it possible to pass in the name of the branch you wish to merge
to. I've defaulted it to "master" for now, but I expect in time this
will change.

### DIFF
--- a/bin/git-merge-branch
+++ b/bin/git-merge-branch
@@ -11,7 +11,7 @@ else
 
   to set the reference and then try merging again."
 
-  exit
+  exit 1
 fi
 
 git fetch origin
--- a/bin/git-merge-branch
+++ b/bin/git-merge-branch
@@ -1,14 +1,20 @@
 #!/bin/sh
 
+main_branch="master"
+
+if [ $# -gt 0 ]; then
+  main_branch=$1
+fi
+
 set -e
 
 git fetch origin
-line_count=$(git diff origin/master..master | wc -l)
+line_count=$(git diff origin/$main_branch..$main_branch | wc -l)
 
 if [ $line_count -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


geoffharcourt:
@danbee can you peek at what I did in https://github.com/thoughtbot/dotfiles/pull/667? I think using `git symbolic-ref` there will make it easier for people to use `base`, `trunk`, `main`, `develop`, or any of the other likely alternatives without pinning to one option. My goal there was to make migration off `master` as hassle-free as possible, I think we can do that here too.

danbee:
@geoffharcourt Thanks for that! I wasn't aware there was any way to get the name of the main branch in git. I've added this in, mostly by just copying and pasting your code with a few changes!

bin/git-merge-branch
```diff
@@ -2,13 +2,25 @@
...
+if git symbolic-ref --short refs/remotes/origin/HEAD >/dev/null; then
+  main_branch="$(git symbolic-ref --short refs/remotes/origin/HEAD | sed 's@^origin/@@')" "$@"
+else
+  echo "You don't have a primary branch reference set for your origin remote.
+  Use:
+  git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/{name_of_your_primary_branch}
+
+  to set the reference and then try merging again."
+
+  exit
```
geoffharcourt, line 14: Should this `exit 1` as well in case someone's grouped operations together with `&&`?

danbee: Good catch! I don't shell script very often...

PR #673, opened 2020-06-19
Initialize new repositories with a main branch

jferris:
New Git repositories start out without any branches, but the HEAD
reference points to "master" by default. This means that the first
commit will create a master branch.

This change adds a HEAD file to the Git template with a different ref.
This means that new repositories will commit to a "main" branch by
default instead.

The HEAD file is added to COPY_ALWAYS, because Git's HEAD must be a
regular file and not a symbolic link.

### DIFF
--- a/git_template/HEAD
+++ b/git_template/HEAD
@@ -1,0 +1,1 @@
+ref: refs/heads/main
--- a/rcrc
+++ b/rcrc
@@ -1,2 +1,3 @@
 EXCLUDES="README*.md LICENSE"
 DOTFILES_DIRS="$HOME/dotfiles-local $HOME/dotfiles"
+COPY_ALWAYS="git_template/HEAD"


tute:
Heya! 👋 

While installing laptop and dotfiles on a new mac I couldn't install asdf. Trying to install it or add plugins would exit early with a `fatal: remote-curl: fetch attempted without a local repo` error message. Also, in `~/.asdf` it would find no repository; I would `rm -rf .git/ && git init` and it _still_ wouldn't find a valid git repository.

After some debugging I found this commit as difference with my other setups. Removing the file made it work again. Maybe there's another way to set the default that doesn't affect other repos. I don't exactly know yet how this two interfere.

tute:
I didn't yet find the root cause of this, but for `rcup` to run vim's `PlugInstall` I had to comment out `templatedir` in `gitconfig` (`  # templatedir = ~/.git_template`).

geoffharcourt:
Possibly relevant as a way to fix this less disruptively going forward: https://github.blog/2020-07-27-highlights-from-git-2-28/#introducing-init-defaultbranch

Cc: @jferris @tute

jferris:
@tute Hi! I'll try to take a look later this week. Hopefully the new `defaultBranch` setting doesn't cause the same issue.

StephenBrown2:
It should not:

> Here, it’s important to note that:
> 
> 1. This configuration variable can be set by the user, and overriding the default value is as easy as:
> 
>     $ git config --global init.defaultBranch main
> 
> 2. This configuration variable *only affects new repositories*, and does not cause branches in existing projects to be renamed. `git clone` will also **continue to respect the HEAD of the repository you’re cloning from**, so you won’t see a change in branch names until a maintainer initiates one.

jferris:
I've opened #676 to use `defaultBranch`. Thanks @tute @geoffharcourt @StephenBrown2 !

tute:
Thank you too!

PR #675, opened 2020-07-02
Set filetype for `typescriptreact` to `typescript`

StevenXL:
As of vim v8.1.1930 (and nvim v0.4.0), *.tsx files get a default filetype of typescriptreact. Many plugins do not recognize `typescriptreact`.

This changes the filetype back to `typescript` from `typescriptreact`. 

I am making this PR to discuss, not necessarily to merge. I do not know the full repercussions of this change, or if there is more elegant way to handle this.

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -50,6 +50,7 @@ augroup vimrcEx
   autocmd BufRead,BufNewFile aliases.local,zshrc.local,*/zsh/configs/* set filetype=sh
   autocmd BufRead,BufNewFile gitconfig.local set filetype=gitconfig
   autocmd BufRead,BufNewFile tmux.conf.local set filetype=tmux
+  autocmd BufRead,BufNewFile *.tsx set filetype=typescript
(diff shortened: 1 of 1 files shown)


geoffharcourt:
Thanks for flagging this!

Some plugins like Coc I think won't handle this change, so I think we need to decide which audience we want to leave out in the cold. Maybe Coc folks will already have this set, so we can assume they will override?

croaky:
I bumped into this awhile back and ended up forcing the syntax only: https://github.com/croaky/laptop/commit/b6ea306d84f5276f421205082224f5b34c76d4dc

StevenXL:
> I bumped into this awhile back and ended up forcing the syntax only: [croaky/laptop@b6ea306](https://github.com/croaky/laptop/commit/b6ea306d84f5276f421205082224f5b34c76d4dc)

Fantastic solution! Thank you.

geoffharcourt:
The power of teamwork!

PR #676, opened 2020-08-28
Use defaultBranch instead of git_template

jferris:
Git 2.28 now has a setting to change the default branch for new
repositories. This sets init.defaultBranch to main and removes the HEAD
file from the git template.

The HEAD file will need to be removed from ~/.git_template manually, as
rcup does not remove deleted files.

https://github.blog/2020-07-27-highlights-from-git-2-28/#introducing-init-defaultbranch

### DIFF
--- a/hooks/post-up
+++ b/hooks/post-up
@@ -8,9 +8,16 @@ else
   curl -fLo "$HOME"/.vim/autoload/plug.vim --create-dirs \
       https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 fi
+
 vim -u "$HOME"/.vimrc.bundles +PlugUpdate +PlugClean! +qa
 
 reset -Q
+
+if [ -f "$HOME/.git_template/HEAD" ] && \
+  [ "$(cat "$HOME/.git_template/HEAD")" = "ref: refs/heads/main" ]; then
+  echo "Removing ~/.git_template/HEAD in favor of defaultBranch" >&2
+  rm -f ~/.git_template/HEAD
+fi
 
 # detect old OS X broken /etc/zshenv and suggest rename
 if grep -qw path_helper /etc/zshenv 2>/dev/null; then
--- a/git_template/HEAD
+++ b/git_template/HEAD
@@ -1,1 +1,0 @@
-ref: refs/heads/main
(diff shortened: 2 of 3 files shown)


mike-burns:
Heck yeah!

Will it be a problem that `HEAD` is in the git_templates dir? We could add a rcm hook if so.

jferris:
@mike-burns There was a bug reported in #673 about the `HEAD` file:

> While installing laptop and dotfiles on a new mac I couldn't install asdf. Trying to install it or add plugins would exit early with a fatal: remote-curl: fetch attempted without a local repo error message. Also, in ~/.asdf it would find no repository; I would rm -rf .git/ && git init and it still wouldn't find a valid git repository.

I checked locally and having `HEAD` doesn't conflict with having `defaultBranch`, but we'll need to remove `HEAD` from the Git template to resolve that issue.

What would an rcm hook look like?

mike-burns:
We already have [a `post-up` hook](https://github.com/thoughtbot/dotfiles/blob/master/hooks/post-up) in this repo, so we'd add to there. It's a simple executable that runs after `rcup` completes.

So we could add this to the end:

```sh
if [ "$(cat ~/.git_template/HEAD)" = "ref: refs/heads/main" ]; then
  echo "removing ~/.git_template in favor of defaultBranch"
  rm -f ~/.git_template/HEAD
fi
```

jferris:
@mike-burns I added a hook in a94ed3a. Works okay for me locally; mind giving it a shot?

hooks/post-up
```diff
@@ -8,10 +8,16 @@ else
   curl -fLo "$HOME"/.vim/autoload/plug.vim --create-dirs \
       https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 fi
-vim -u "$HOME"/.vimrc.bundles +PlugUpdate +PlugClean! +qa
+vm -u "$HOME"/.vimrc.bundles +PlugUpdate +PlugClean! +qa
```
mike-burns, line 11: What happened here?

mike-burns (approved):
Works for me. LGTM (though I think you might have typoed a line elsewhere in post-up).

hooks/post-up
jferris: Uhhh, I don't have that change locally. Wild. I just re-pushed my commit and had to force push. Looks like it's not there now?

mike-burns: Computers are such a great idea.

PR #691, opened 2021-12-03
fix: stock vim on Catalina does not support diffopt+=vertical

barretlee:


mikegee:
👋  apparently this similar solution is more correct: https://github.com/thoughtbot/dotfiles/issues/655#issuecomment-605019271

I don't really understand these things. I just remember seeing this error mentioned in a PR or issue before and found it. :) good luck

PR #693, opened 2022-01-14
Add helper to list process running in a given port

aledustet:
Why:
----

Whenever we run a command to use a certain port and we get an error that the port is in use, we have can either:

1. Clear the port so our new process can use it
2. Try and use a new port for the command we just typed

In order to make a decision about it, we need information on what is running on that port, so we can determine if we want to stop that process or if it's important enough to keep.

What:
----

Two new functions are added for our `zsh` customizations. One to output information about the process using a provided port:

``` bash
whats_in_port 3000
COMMAND   PID   USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
ruby    25583   root   11u  IPv4 0xee20607697a79bf7      0t0  TCP *:irdmi (LISTEN)
```

This command uses the [lsof] program to get information of the process utilising that port.

And another one making use of the previous command to kill the process running in that port:

``` bash
clear_port 3000
```

[lsof]: https://linux.die.net/man/8/lsof

### DIFF
--- a/bin/clear-port
+++ b/bin/clear-port
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# Kills the process running on the provided port
+#
+# clear-port 3000
+
+if [ -n "$1" ]; then
+  port_num="$(lsof -ti4TCP:"$1")"
+  if [ $? -eq 0 ]; then
+    kill "$port_num"
+  fi
+else
+  echo >&2 Usage: clear_port port-number
+  exit 1
+fi
--- a/bin/whats-in-port
+++ b/bin/whats-in-port
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+# List process running on provided port
+#
+# whats-in-port 3000
+#
+# output:
+# COMMAND   PID   USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
+# ruby    25583   root   11u  IPv4 0xee20607697a79bf7      0t0  TCP *:irdmi (LISTEN)
+
+if [ -n "$1" ]; then
+  lsof -ni4TCP:"$1"
+else
+  echo >&2 Usage: whats_in_port port-number
+  exit 1
+fi


zsh/functions/whats-in-port
```diff
@@ -0,0 +1,23 @@
+# List process running on provided port
+#
+# whats_in_port 3000
+#
+# output:
+# COMMAND   PID   USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
+# ruby    25583   root   11u  IPv4 0xee20607697a79bf7      0t0  TCP *:irdmi (LISTEN)
+
+whats_in_port() {
```
MatheusRich, line 9: THANK ~GOD~ ALEJANDRO!

zsh/functions/whats-in-port
```diff
@@ -0,0 +1,23 @@
...
+  if [ "$1" ]; then
+    lsof -n -i4TCP:"$1"
+    return
+  fi
+  echo "Port number needed to determine accessing process"
```
mike-burns, line 14: This should probably be an error,

```suggestion
  echo >&2 "Port number needed to determine accessing process"
  exit 1
```

zsh/functions/whats-in-port
```diff
@@ -0,0 +1,23 @@
...
+}
+
+# Kills the process running on the provided port
+#
+# clear_port 3000
+
+clear_port() {
+  kill "$(whats_in_port "$1")"
```
mike-burns, line 22: How does this work?

```
~% lsof -n -i4TCP:3000
COMMAND    PID USER   FD   TYPE  DEVICE SIZE/OFF NODE NAME
ruby    315558 mike   12u  IPv4 8405941      0t0  TCP 127.0.0.1:3000 (LISTEN)
```

That doesn't look like what I pass to `kill(1)`.

MatheusRich (approved):

mike-burns (commented):
Why is this a function? What do you think of making it a program in `bin`, available to all our shell users and not just Zsh users?

zsh/functions/whats-in-port
stevehanson: I think this is missing the `-t` parameter:

```
~% lsof -n -i4TCP:3000 -t
31558
```

edwardloveall: The [-t flag](https://linux.die.net/man/8/lsof) might be helpful here:

> -t
>    This option specifies that lsof should produce terse output with process identifiers only and no header - e.g., so that the output may be piped to **kill**(1). This option selects the -w option.

zsh/functions/whats-in-port
stevehanson: I have a similar [function in my dotfiles](https://github.com/stevehanson/dotfiles/blob/main/zsh/functions/ports). Mine has a default argument of 3000 instead of erroring, but I think I like that yours fails if no argument is provided. 3000 makes sense for me but maybe not for other users, so it'd be hard to find a perfect default.

stevehanson (commented):
Nice! I have variations of these in my dotfiles and find them very useful.

zsh/functions/whats-in-port
aledustet: Yup great catch everyone, I had this function on another file and decided against including it 😅

zsh/functions/clear_port
```diff
@@ -0,0 +1,11 @@
+# Kills the process running on the provided port
+#
+# clear_port 3000
+
+clear_port() {
+  if [ "$1" ]; then
+    kill "$(lsof -t -i4TCP:"$1")"
+    return
+  fi
+  echo >&2 "Port number needed to determine process id"
```
mike-burns, line 10: We should probably pass failure back to the shell (`exit 1`) so it knows it's an error, both here and in `whats_in_port`.

We can get fancier, too: don't kill if we can't find it, like so:

```sh
if [ -n "$1" ]; then
  port_num="$(lsof -ti4TCP:"$1")"
  if [ $? -eq 0 ]; then
    kill "$port_num"
  fi
else
  echo >&2 Usage: clear_port port-number
  exit 1
fi
```

edwardloveall: So I tried to add this, but it ends up closing my current terminal tab when I add the `exit 1`. Is there a way to do this that doesn't do that?

mike-burns: Yes:

> Why is this a function? What do you think of making it a program in bin, available to all our shell users and not just Zsh users?

mike-burns: Though if you want to keep this a function you could probably use `return` instead of `exit`. Though I don't know why this should be a function.

edwardloveall: Ah I missed that. What does "program in bin" mean?

mike-burns: Move these two files to `bin/`, add `#!/bin/sh` to the top, and remove the function declaration around the body (so remove `clear_port() {` and `whats_in_port() {`).

Shell functions aren't very useful but if you need e.g. a program that will exit your shell or change directories on your behalf then they're quite handy. They're incredibly powerful, though, at the expense of being non-portable and bug-prone.

edwardloveall: Ah I see and then they'll be in the path and can be invoked similarly. I was missing the forest through the trees, thanks!

edwardloveall: Done 12cf89a

mike-burns (approved):
woot! Thank you for this.

PR #695, opened 2022-03-16
Updated README to use encrypted git connection

fa11enangel:
Due to current changes on GitHub, users are forced to use encrypted
connections only, see:
https://github.blog/2021-09-01-improving-git-protocol-security-github/

This is the corresponding change in READMEs.


geoffharcourt (approved):
Looks great, thanks for updating this!

PR #700, opened 2022-05-08
Uses #!/bin/bash instead of #!/bin/sh for bin/ scripts

dorianmariefr:
I notice vim's syntax highlighting was wrong for
some bin/ scripts

It was because #!/bin/sh doesn't necessarily support
the $() syntax

As recommended in the thoughtbot's guides:

> Don't use a /bin/sh shebang unless you plan to test
> and run your script on at least: Actual Sh, Dash in
> POSIX-compatible mode (as it will be run on Debian),
> and Bash in POSIX-compatible mode
> (as it will be run on macOS).

https://github.com/thoughtbot/guides/tree/main/shell

### DIFF
--- a/bin/bundler-search
+++ b/bin/bundler-search
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Search your bundle for the provided pattern
 #   Requires bundler 1.8+ for execution as a bundler subcommand.
--- a/bin/clear-port
+++ b/bin/clear-port
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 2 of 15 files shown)


dorianmariefr:
Here is what I was seeing:

Screen Shot 2022-05-08 at 11 30 04

mike-burns:
That's because you are using non-POSIX mode in vim. Here's the fix:

```vi
let g:is_posix = 1
```

I don't have GNU Bash installed on most of my computers so I'd rather we don't use it for our dotfiles.

dorianmariefr:
Ah okay, I assumed every computer has bash nowadays but that's not true

Thanks for the suggested fix

Issue #702, opened 2022-07-27
zsh profile not fully loading on terminal startup (iTerm2 + Monterrey)


markelld:
Hello,

This might be a simple oversight on my part, but I noticed I've had to manually run `source ~/.zshrc` each time I start my terminal, in order to access commands like `node` and `yarn`.

Dotfiles were installed on a new Macbook Air 2022, running Monterrey 12.4 (fresh install).

<img width="395" alt="Screen Shot 2022-07-27 at 3 47 47 PM" src="https://user-images.githubusercontent.com/60442714/181369458-74b2c02a-5585-4181-b829-7f6ecc74b3a6.png">

Here is my `$PATH` before and after running `source`:
<img width="1463" alt="Screen Shot 2022-07-27 at 3 40 22 PM" src="https://user-images.githubusercontent.com/60442714/181369452-425eefd8-f10f-4842-bbb5-1aec0f9ec9a7.png">

I ran followed the instructions on the README, but seems like an additional step is required? 
Could it be a preference/setting in iTerm2?

Thanks in advance.

danielnolan:
Hi @markelld thanks for the report. Could you try creating a `.zprofile` in your home directory with the following entry
`eval "$(/opt/homebrew/bin/brew shellenv)"` and see if that resolves your path issue? 
I recently setup an M1 mac and had a similar issue until I added `.zprofile` with this command.

markelld:
@danielnolan  So I tried this command but it has not rectified the issue, sorry for the late reply. Also I am getting not output from the command.

danielnolan:
@markelld did you try @mrjonesbot suggestion? Is this still an issue for you?

PR #707, opened 2022-09-20
Add zprofile with eval command for homebrew

danielnolan:
* I recently setup my M1 mac and had a similar problem to this issue
  reported here https://github.com/thoughtbot/dotfiles/issues/702
  The fix for me was to do what homebrew says after the install which is
  to add to or create a .zprofile with the following command
  `eval "$(/opt/homebrew/bin/brew shellenv)"`
* Introduce a zprofile with commands to eval homebrew on mac or linux if
  the homebrew directory exists

### DIFF
--- a/zprofile
+++ b/zprofile
@@ -1,0 +1,7 @@
+if [ -d "/opt/homebrew" ]; then
+  eval "$(/opt/homebrew/bin/brew shellenv)"
+elif [ -d "~/.linuxbrew" ]; then
+  eval "$(~/.linuxbrew/bin/brew shellenv)"
+elif [ -d "/home/linuxbrew" ]; then
+  eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+fi


zprofile
```diff
@@ -0,0 +1,5 @@
+if [ -d "/opt/homebrew" ]; then
+  eval "$(/opt/homebrew/bin/brew shellenv)"
+elif [[ -d /home/linuxbrew ]]; then
```
thoughtbot-summer, line 3: Should we use the same syntax as above?

```suggestion
elif [ -d "/home/linuxbrew" ]; then
```

thoughtbot-summer, line 3: Homebrew on Linux [may also install to `$HOME/.linuxbrew`](https://docs.brew.sh/Homebrew-on-Linux#install):

> The installation script installs Homebrew to /home/linuxbrew/.linuxbrew using sudo if possible and within your home directory at ~/.linuxbrew otherwise.

Can we check for that as well?

mike-burns:
Curious about `.zprofile`.

The [Zsh docs](https://zsh.sourceforge.io/Intro/intro_3.html) say

> `.zlogin' is sourced in login shells. It should contain commands that should be executed only in login shells. `.zlogout' is sourced when login shells exit. `.zprofile' is similar to `.zlogin', except that it is sourced before `.zshrc'. `.zprofile' is meant as an alternative to `.zlogin' for ksh fans; the two are not intended to be used together, although this could certainly be done if desired. `.zlogin' is not the place for alias definitions, options, environment variable settings, etc.; as a general rule, it should not change the shell environment at all. Rather, it should be used to set the terminal type and run a series of external commands (fortune, msgs, etc).

When I [look at shellenv](https://github.com/Homebrew/brew/blob/cd405c791169782c44d615205f939403431d0389/Library/Homebrew/cmd/shellenv.sh), I see a whole bunch of environment variables.

What do you think of moving this to `zshenv`?

thoughtbot-summer:
@mike-burns If so, maybe it's worth also opening a PR into Homebrew to change that?

mike-burns:
@thoughtbot-summer good idea. Interestingly, it does mention `.zshenv` in a few places in the repo. I think this was a `.bash_login` -> Zsh confusion (combined with the fact that Zsh has too many config files...).

Anyway, I'll see if I can get a PR open today.

mike-burns:
https://github.com/Homebrew/brew/pull/13920 - let a thousand unix nerds bloom.

danielnolan:
@mike-burns I think the use of `.zprofile` might have to do with apple and the path_helper included in `/etc/zprofile` which is loaded after `~/.zshenv` and can clobber the path as discussed here https://apple.stackexchange.com/questions/432226/homebrew-path-set-in-zshenv-is-overridden. 
I did the same experiment on my mac as one of the comments on that stack exchange post and you can see that the `/etc/zprofile`  is sourced after `~/.zshenv`.  

Screen Shot 2022-09-23 at 4 09 04 PM

mike-burns:
In light of @danielnolan 's research, I guess `.zprofile` is the right file. LGTM.

danielnolan:
@thoughtbot-summer I have made the changes you suggested, please have a look when you can :)

thoughtbot-summer (approved):
Looks great! :sparkling_heart:

PR #709, opened 2022-11-03
Git global config to not ignore case

cpytel:
Originally a suggestion by @ACPK here https://github.com/thoughtbot/laptop/issues/601, after discussing it we thought the dotfiles would be a better place for this change.

This change ensures that capitalization change in filenames are tracked via Git, and original case is preserved.

### DIFF
--- a/gitconfig
+++ b/gitconfig
@@ -16,6 +16,7 @@
 [core]
   excludesfile = ~/.gitignore
   autocrlf = input
+  ignorecase = false
 [merge]
   ff = only
 [commit]


danielnolan (approved):
👍

mike-burns:
This makes sense for APFS. Quoting the docs:

> if a directory listing finds "makefile" when Git expects "Makefile", Git will assume it is really the same file, and continue to remember it as "Makefile".

On the other hand, on FFS, UFS, ext2, etc., I'd find that a little surprising.

The docs also say:

> The default is false, except git-clone(1) or git-init(1) will probe and set core.ignoreCase true if appropriate when the repository is created.

And sure enough, on an APFS that I have available:

```
~% grep ignorecase .dotfiles/.git/config
	ignorecase = true
```

Do we need to set it globally?

stevehanson (commented):
This is interesting. I am on MacOS and do not have any global config for `ignorecase`. 

I used @mike-burns command to check the local settings of my current project and found that `ignorecase = true` was set (presumably automatically when I cloned the repo) and that case changes were not picked up:

```shell
> grep ignorecase .git/config 
        ignorecase = true
> mv Gemfile gemfile
> git status
  nothing to commit, working tree clean
```

I then set the global config: `git config --global core.ignorecase false`, and then cloned the repo fresh, but the results were the same: `ignorecase = true` was set in the local `.git/config` and file name changes were not picked up.

The only way I was able to get my repo to recognize case changes was by running `git config --local core.ignoreCase false`, which removed the `ignorecase = true` line from `.git/config`.

After this experiment, I'm wondering what effect the global config can have?

danielnolan:
> Git relies on the proper configuration of this variable for your operating and file system. Modifying this value may result in unexpected behavior.
After reading this line in the docs, I think it's best we just leave it alone. It's configured by git and should ignore case on case insensitive files systems.

cpytel:
Thanks all. Based on this discussion, I guess I'm going to close this issue.

There still seems to be something to this, but I don't know enough to have a strong opinion otherwise. From my perspective, it seems like it is better to have everyone working on a project using the same case for the files in git. I also recently had an issue where that wasn't the case and it was very confusing.

PR #710, opened 2022-12-12
Fix sourcing of asdf.sh

abhineet97:
Installing asdf via Homebrew now requires sourcing from a different path.

This same issue was recently fixed in thoughtbot/laptop by @cpytel, and so I've copied the changes from this commit: https://github.com/thoughtbot/laptop/commit/aa3a84e15e76af9bf3713c350354bc42e59749bc


jasquat:
@abhineet97, @cpytel  this is invalid zsh syntax and cannot be loaded. The elif line (line 8) needs to be `elif which brew >/dev/null; then`.

abhineet97:
You're right, @jasquat! Yikes 😬

PR #713, opened 2023-03-08
Update the check for fzf installed via homebrew

### LINKED ISSUE #681: New location for fzf under macOS with ARM architecture
On macOS with M1 chip, Homebrew is now installed by default in `/opt/homebrew/` (see https://github.com/Homebrew/install/pull/373).

The current installation process determines whether fzf was installed via Homebrew, but only checks under `/usr/local/`: https://github.com/thoughtbot/dotfiles/blob/2a59c1890f81bffc1db79cae4482ce2b706b0f79/vimrc.bundles#L20-L26

I think adding a new check for `/opt/homebrew/` might help for machines with M1 chips.

danielnolan:
* The existing check for fzf installed via homebrew doesn't work for
  Apple silicon Macs, nor does it work for those who use homebrew on
  Linux.
* Add brew_fzf_path variable to get the brew prefixed path
* Update the isdirectory conditional to check this variable and use it for
  the Plug function call if it is a directory
* Closes #681

### DIFF
--- a/vimrc.bundles
+++ b/vimrc.bundles
@@ -19,8 +19,11 @@ Plug 'christoomey/vim-run-interactive'
 
 " If fzf has already been installed via Homebrew, use the existing fzf
 " Otherwise, install fzf. The `--all` flag makes fzf accessible outside of vim
-let g:brew_fzf_path = trim(system("brew --prefix fzf"))
-if isdirectory(g:brew_fzf_path)
+if executable("brew")
+  let g:brew_fzf_path = trim(system("brew --prefix fzf"))
+endif
+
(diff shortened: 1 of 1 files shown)


danielnolan:
@mikegee I made a small change to ensure that brew is executable. If fzf isn't installed via homebrew Plug installs it to `~/.fzf` https://github.com/thoughtbot/dotfiles/pull/713/files#diff-d38c94eb3c98eb2f3fbf6ced6c8e3c36a0135e5ad558a10d09efc83f31fc048aR29

mikegee (approved):
Seems legit

PR #719, opened 2023-05-26
Let users to use the arrow keys

neilvcarvalho:
During pairing sessions, we encounter difficulties when collaborating
with non-Vim users due to the dotfiles blocking the use of arrow keys,
as `h`, `j`, `k` and `l` motions are unfamiliar to non-Vim users. While
those motions are undoubtedly more productive than arrow keys, blocking
them makes collaboration worse.

Users who want to force themselves to stop using the arrow keys can
add the key mappings to their own local dotfiles, or use a Vim plugin
that includes similar behavior.

### DIFF
--- a/vimrc
+++ b/vimrc
@@ -127,12 +127,6 @@ inoremap <S-Tab> <C-n>
 " Switch between the last two files
 nnoremap <Leader><Leader> <C-^>
 
-" Get off my lawn
-nnoremap <Left> :echoe "Use h"<CR>
-nnoremap <Right> :echoe "Use l"<CR>
-nnoremap <Up> :echoe "Use k"<CR>
-nnoremap <Down> :echoe "Use j"<CR>
-
 " vim-test mappings
 nnoremap <silent> <Leader>t :TestFile<CR>
 nnoremap <silent> <Leader>s :TestNearest<CR>


jferris (approved):

geoffharcourt (approved):
Have had the same issue

mike-burns:
Should we just, like, allow arrow keys without any warning?

rtaibah:
In my opinion blocking arrow keys is for beginners who are serious about learning vim. I blocked them when I first started over a decade ago, and it literally just served me for a couple of days. By then I got rid of the habit, and the config stayed there ever since.

It came back to bite me when I was pairing with @rakeshpetit a few weeks ago.

Also quite frankly, the config seemed 'holier than thou' to me back then, but I went along with it.

I guess what I am trying to say is: just remove it. A beginner most probably won't be using these files, and it just gets in the way of pairing.

rtaibah:
> But I don't agree with this opinion that someone using arrow keys is a beginner.

Sorry Rakesh, I didn't mean it in a bad way or diminishing fashion. I just meant that blocking arrow keys is a "general wisdom" that is recommended for people who want to pickup vim.

mike-burns:
For clarity, the reason arrow keys are considered bad for Vim beginners is because there are very few times you actually want to go one character left, right, up, or down. In that sense, hjkl on their own are also bad form. Putting a command in front (`dl`, `2cj`, etc) is a more common style.

All that said, I'd rather we ship a minimal Vim config.

(Also, I switched to Neovim so I don't have much of a dog in this fight.)

PR #726, opened 2023-08-04
Add git co-upstream-pr command

stefannibrasil:
When reviewing Pull Requests from open source projects, it's handy to have a way to just pass the PR number and the local branch name.

With this new command, to fetch a PR from an upstream remote repo, we just need the PR number and the name of the local branch:

$ co-upstream-pr 001 test_generator_pr

git will checkout the upstream PR in the local test_generator_pr branch.

If there is no upstream remote set, we get a nice message guiding users to set it first:

```
You don't have an upstream remote set.
Use:
  git remote add upstream {upstream_remote_url}
  to set the reference and then try again.
```

Co-authored by @minaslater 🎉

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,7 @@ configuration:
 
 [git](http://git-scm.com/) configuration:
 
+- Adds a `co-upstream-pr $PR_NUMBER $LOCAL_BRANCH_NAME` alias to checkout remote upstream branch into a local branch.
 - Adds a `create-branch` alias to create feature branches.
 - Adds a `delete-branch` alias to delete feature branches.
 - Adds a `merge-branch` alias to merge feature branches into master.
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ configuration:
 
 [git](http://git-scm.com/) configuration:
 
-- Adds a `co-upstream-pr $PR_NUMBER $LOCAL_BRANCH_NAME` alias to checkout remote upstream branch into a local branch.
+- Adds a `co-upstream-pr $PR_NUMBER $LOCAL_BRANCH_NAME` subcommand to checkout remote upstream branch into a local branch.
 - Adds a `create-branch` alias to create feature branches.
 - Adds a `delete-branch` alias to delete feature branches.
 - Adds a `merge-branch` alias to merge feature branches into master.
--- a/bin/git-co-upstream-pr
+++ b/bin/git-co-upstream-pr
@@ -5,13 +5,15 @@ set -e
 pull_request_number=$1
 local_branch_name=$2
 
-if git remote -v | grep upstream; then
+if git remote -v | grep -q upstream; then
   git fetch upstream "pull/$pull_request_number/head:$local_branch_name"
   git checkout "$local_branch_name"
 else
-  echo "You don't have an upstream remote set.
-  Use:
+  cat <<HELP
+You don't have an upstream remote set.
+Use:
   git remote add upstream {upstream_remote_url}
 
-  to set the reference and then try again."
+to set the reference and then try again.
+HELP
 fi
--- a/bin/git-co-upstream-pr
+++ b/bin/git-co-upstream-pr
@@ -5,6 +5,11 @@ set -e
 pull_request_number=$1
 local_branch_name=$2
 
+if [ -z "$pull_request_number" -o -z "$local_branch_name" ]; then
+  echo "usage: git co-upstream-pr <pull_request_number> <local_branch_name>"
+  exit 1
+fi
+
 if git remote -v | grep -q upstream; then
   git fetch upstream "pull/$pull_request_number/head:$local_branch_name"
   git checkout "$local_branch_name"
--- a/bin/git-co-upstream-pr
+++ b/bin/git-co-upstream-pr
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+set -e
+
+pull_request_number=$1
+local_branch_name=$2
+
+if git remote -v | grep upstream; then
+  git fetch upstream "pull/$pull_request_number/head:$local_branch_name"
+  git checkout "$local_branch_name"
+else
+  echo "You don't have an upstream remote set.
+  Use:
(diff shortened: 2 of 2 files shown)


bin/git-co-upstream-pr
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -e
+
+pull_request_number=$1
+local_branch_name=$2
+
+if git remote -v | grep upstream; then
```
mike-burns, line 8: I _think_ this will also print the word `upstream` to the screen, which you can avoid like so:

```suggestion
if git remote -v | grep -q upstream; then
```

bin/git-co-upstream-pr
```diff
@@ -0,0 +1,17 @@
...
+  git fetch upstream "pull/$pull_request_number/head:$local_branch_name"
+  git checkout "$local_branch_name"
+else
+  echo "You don't have an upstream remote set.
+  Use:
+  git remote add upstream {upstream_remote_url}
```
mike-burns, line 14: Is `upstream` different from `origin`? Maybe I don't know how to use this command.

bin/git-co-upstream-pr
```diff
@@ -0,0 +1,17 @@
...
+set -e
+
+pull_request_number=$1
+local_branch_name=$2
+
+if git remote -v | grep upstream; then
+  git fetch upstream "pull/$pull_request_number/head:$local_branch_name"
+  git checkout "$local_branch_name"
+else
+  echo "You don't have an upstream remote set.
```
mike-burns, line 12: Up to you, but two of the things you're doing are unreliable (but not necessarily in an important way):

- `echo` is considered fragile unless you pass very, very boring things to it. You're passing a boring thing, but it could be more boring. Makes me nervous but it's probably fine.
- Multi-line quotes have a fraught history. Again, it will probably work on modern computers, but it makes me nervous.

Here's another idea:

```sh
cat <<HELP
You don't have an upstream remote set.
Use:
  git remote add upstream {upstream_remote_url}
to set the reference and then try again.
HELP
```

This uses a heredoc (`here-doc`, according to one manpage I read), which has more support across shells, and sends the output of that straight into `cat`.

README.md
```diff
@@ -198,6 +198,7 @@ configuration:
 
 [git](http://git-scm.com/) configuration:
 
+- Adds a `co-upstream-pr $PR_NUMBER $LOCAL_BRANCH_NAME` alias to checkout remote upstream branch into a local branch.
```
mike-burns, line 201: I believe this is a subcommand, not an alias.

```suggestion
- Adds a `co-upstream-pr $PR_NUMBER $LOCAL_BRANCH_NAME` subcommand to checkout remote upstream branch into a local branch.
```

bin/git-co-upstream-pr
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -e
+
+pull_request_number=$1
+local_branch_name=$2
```
mike-burns, line 6: If these aren't passed, we should probably error helpfully:

```sh
if [ -z "$pull_request_number" -o -z "$local_branch_name" ]; then
  echo "usage: git co-upstream-pr <pull_request_number> <local_branch_name>"
  echo 1
fi
```

mike-burns (commented):
Nice!

The thoughtbot `laptop` script installs the `gh` command, which has a `pr` subcommand:

```sh
gh pr checkout 353 # check out PR #353 into a local branch
```

This seems to do a similar thing. Do we want both?

stefannibrasil:
> The thoughtbot `laptop` script installs the `gh` command, which has a `pr` subcommand:
> 
> ```shell
> gh pr checkout 353 # check out PR #353 into a local branch
> ```
> 
> This seems to do a similar thing. Do we want both?

It does a similar thing, yes. But it checkouts the PR into a local branch from a remote upstream (forked) repository. For example, if I want to check out [this PR](https://github.com/faker-ruby/faker/pull/2794) from faker-ruby, I can't use `gh pr checkout 2794` because I don't have that reference in my local git. To checkout that branch, I need to run the commands added here in this PR.

bin/git-co-upstream-pr
stefannibrasil: Yes, they are different! `upstream` repositories are the original repository that a fork is created from. GitHub has some details here: https://docs.github.com/en/get-started/quickstart/fork-a-repo#configuring-git-to-sync-your-fork-with-the-upstream-repository

I have mainly used `upstream` remote repos for forked open source projects. But they can be used for any repos where you want to have your own fork, but still wants to fetch the original repo's updates.

bin/git-co-upstream-pr
stefannibrasil: I tried with both and didn't see any difference. I guess it doesn't hurt adding it. Thank you!

bin/git-co-upstream-pr
stefannibrasil: Oh, I didn't know that! I was following the existing format from the other commands. I do not want to do unreliable things :O

Thank you for this lesson and suggestion, Mike!

Changed this in d2069f516a62915b6f7e4c2d84dbbcee096247f3

README.md
stefannibrasil: Changed it in be49b1df70e023cd80569cf53ffd4c9173d0e17b

bin/git-co-upstream-pr
stefannibrasil: OH, good idea! This entire script writing is so new to me, this totally escaped me. Added this check and instead of `echo 1`, I think we can exit here instead of proceeding to the next check. Did that in e0d610aba0edd1bc80862325a015f81856fb542b

mike-burns (approved):
Ooh, I get it now! Thank you for patiently explaining this command to me.

LGTM!

stefannibrasil:
> Ooh, I get it now! Thank you for patiently explaining this command to me.
> 
> LGTM!

Thank you for the review and the mentoring ❤️

PR #728, opened 2023-10-06
Introduce `railsrc`

stevepolitodesign:
Introduces [railsrc][1] in an effort to prepare for the eventual [removal of the
suspenders executable][2].

With this change, calling `rails new` will automatically set the following
options:

```
--database=postgresql
```

If you need to use another database, you can override this value like so: `rails
new my_app --database=mysql` or `rails new my_app --no_rc` to skip this file all
together

[1]: https://github.com/rails/rails/blob/7f7f9df8641e35a076fe26bd097f6a1b22cb4e2d/railties/lib/rails/generators/rails/app/USAGE#L5C1-L7
[2]: https://github.com/thoughtbot/suspenders/blob/main/GOALS.md#from-here-to-there

### DIFF
--- a/railsrc
+++ b/railsrc
@@ -1,0 +1,1 @@
+--database=postgresql
--- a/README.md
+++ b/README.md
@@ -214,6 +214,14 @@ configuration:
 - Add trusted binstubs to the `PATH`.
 - Load the ASDF version manager.
 
+[Rails](https://rubyonrails.org)
+
+- Adds [railsrc][] with `--database=postgresql`. If you need to use another
+  database, you can override this value like so: `rails new my_app --database=mysql` or
+`rails new my_app --no_rc`
+
+[railsrc]: https://github.com/rails/rails/blob/7f7f9df8641e35a076fe26bd097f6a1b22cb4e2d/railties/lib/rails/generators/rails/app/USAGE#L5C1-L7
+
 Shell aliases and scripts:
 
 - `b` for `bundle`.


railsrc
```diff
@@ -0,0 +1,2 @@
+--database=postgresql
+--css=tailwind
```
stevepolitodesign, line 2: How do we feel about this?

JoelQ: Is this what we use to initialize most new apps these days? I haven't done `rails new` in a few years 😅

JoelQ:
Would this make it more challenging to generate a Rails app that deviates from this standard? Since now it wouldn't just be picking a different generator or setting a command-line flag but would require modifying a dotfile (or potentially adding a local `.railsrc`?)

JoelQ:
Perhaps out of scope for this PR but I wonder if there are other flags we'd want to set here? I know when I've generated new Rails apps I've always set the test framework to RSpec but IIRC @seanpdoyle might be trying to change that convention.

mike-burns:
> Would this make it more challenging to generate a Rails app that deviates from this standard?

I'm pretty sure a command line flag overrides the dotfile.

railsrc
mike-burns: This causes Rails to run `rails css:install:tailwind` after making the app. We can do that ourselves if we want Tailwind, not sure why it needs to be part of `rails new`.

mike-burns: It looks like all the CSS frameworks depend on `css:install`, which might bring in [good stuff](https://github.com/rails/cssbundling-rails/blob/main/lib/install/install.rb). Is that the only way to get `css:install` to run?

brimatteng:
Personally, I don't like Tailwind and I wouldn't reach for it by default. I think especially since you need to also run a build process for Tailwind

railsrc
pangolingo: On the Ignite team, we use Tailwind by default. At least we've used it on the last few projects I was involved in. I would like to hear from @kvnkwon  and @vendelal (both out sick today) if they feel like we should continue to use it going forward.

However Ignite rarely starts a new project in Rails so our opinion may be irrelevant here.

stevepolitodesign:
@JoelQ 

> > Would this make it more challenging to generate a Rails app that deviates from this standard?
> 
> I'm pretty sure a command line flag overrides the dotfile.

Just confirming this is true.  Running `rails new my_app --database=mysql` would override `postgresql`. However, there's no option to "unset" the CSS.

However, there's always the option to run `rails new my_app --no-rc` to skip the `.railsrc` all together.

railsrc
stevepolitodesign: I should point at that right now, [suspenders generates new apps with postcss](https://github.com/thoughtbot/suspenders/blob/55cb5c246fe53aea4cf53436c7c208b40ad13e85/lib/suspenders/generators/app_generator.rb#L34-L35).

It sounds like there's no disagreement about using PostgreSQL as our preferred database, but it does sound like there's a disagreement in how we prefer to style applications.

DoodlingDev: I think because there's disagreement it points to there not being alignment on a standard. In the css instance, personally I feel tailwind or other frameworks should be opt-in. That's because I'm not really a fan, and I know there are at least a handful of others that would rather not work in tailwind given the chance.

stevepolitodesign: I ended up removing `--css=tailwind` based on this discussion. We can consider addressing how we want to handle styles in a [new version of Suspenders](https://github.com/thoughtbot/suspenders/pull/1135) I'm exploring.

mike-burns (approved):

PR #729, opened 2023-10-23
✨ Add an `..` alias for quicker navigation

purinkle:
Before, we would have to use five keystrokes to move up a directory in the terminal. We wasted so much unnecessary time. We added an `..` alias to reduce the keystrokes and gain productivity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ configuration:
 
 Shell aliases and scripts:
 
+- `..` for quicker navigation to the parent directory.
 - `b` for `bundle`.
 - `g` with no arguments is `git status` and with arguments acts like `git`.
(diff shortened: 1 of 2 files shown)


mikegee:
Didn't it work before the alias? There's a `setopt autocd` in these configs already I think.

purinkle:
> Didn't it work before the alias? There's a `setopt autocd` in these configs already I think.

@mikegee , thank you for pointing out the Z shell configuration.

I have used these dotfiles for almost a decade, and today is the first day I learned about this line.

Thank you.

purinkle:
This change is unnecessary because we have a Z shell config that adds the same behaviour.

I will close this pull request and remember to read the manual in the future.

Issue #738, opened 2023-12-15
Sourcing asdf has broken the zsh


lngsx:
Recently, I just set up a Linux, Fedora actually, with no homebrew and had 
this issue. 

![Screenshot from 2023-12-15 22-59-27](https://github.com/thoughtbot/dotfiles/assets/1572632/02cf81c5-d664-42c2-9aeb-122bac903a81)

I suspect the problem comes from this commit. https://github.com/thoughtbot/dotfiles/commit/fc2624997b1439db2024d8f00ddfdb0dcdf24d37

- This code assumes that asdf is already installed in the system, either through
homebrew or regular asdf.
- This code assumes homebrew already installed in the system. 

In a system with no asdf and Homebrew installed, the shell will run into the second case, executing `which brew` which will cause an error.

I'm not sure what's really going on here. The zsh script is too hardcore for
me. But I believe something's broken when you try to execute brew when it doesn't exist in the system.

My naive implementation would be using the same code introduced in this commit [#704.](https://github.com/thoughtbot/dotfiles/commit/fc9f7798f85a56297f48785dd07685d5cbf4a1a6)
to check if Homebrew is present in the system before proceed to source the asdf

Thank you.

lngsx:
Just open the PR to fix this.

PR #739, opened 2023-12-15
Fix sourcing of asdf

lngsx:
From https://github.com/thoughtbot/dotfiles/issues/738

Change the way to check if homebrew is present.
The original code assumed that homebrew was already installed.

This same code was introduced in this commit [#704.](https://github.com/thoughtbot/dotfiles/commit/fc9f7798f85a56297f48785dd07685d5cbf4a1a6) to check if Homebrew is present in the system.

### DIFF
--- a/zsh/configs/post/path.zsh
+++ b/zsh/configs/post/path.zsh
@@ -5,9 +5,7 @@ PATH="$HOME/.bin:/usr/local/sbin:$PATH"
 if [ -f "$HOME/.asdf/asdf.sh" ]; then
   . "$HOME/.asdf/asdf.sh"
 # Checking homebrew existance before loading asdf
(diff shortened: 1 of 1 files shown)


mikegee (commented):
This confuses me. It seems like a good idea to check whether we can run `brew` before running `brew` in the command that follows. Checking for the directories homebrew might be installed in is an indirect way to check for the `brew` command.

The internet seems to think `command -v brew` is preferable to `which brew`. Would you see if that behaves better on your system?

lngsx:
Thank you for looking into this.

I have seen that before on stackoverflow. These shell scripts were too scary for me. There are so many ways to test for the existence of homebrew so I went with the one that was active in the main repo.

Anyway, just make an update to the code following your suggestions, and have tested it on Fedora with and without brew.

PR #740, opened 2024-02-07
Add a script to update Bundler in several projects

smaboshe:
This commit adds a script that uses the GitHub CLI and Docker to update Bundler in several projects (defined in `~/.config/bundler-projects`).

It builds on work by @nickcharlton.

Ref:
- https://github.com/nickcharlton/dotfiles/blob/main/bin/update-bundler
- https://github.com/smaboshe/dotfiles-local/blob/main/bin/update-bundler

### DIFF
--- a/bin/update-bundler
+++ b/bin/update-bundler
@@ -15,7 +15,7 @@ set -e
 
 BUNDLER_VERSION="$1"
 if [ -z "$BUNDLER_VERSION" ]; then
-    echo "Usage: Please provide a Bundler version. For example:"
+    echo "Usage: update-bundler BUNDLER_VERSION"
     echo "update-bundler 2.5.4"
     exit 1
 fi
--- a/bin/update-bundler
+++ b/bin/update-bundler
@@ -1,0 +1,78 @@
+#!/bin/sh
+
+# Use Docker to update Bundler in several projects and open a PR for each one.
+#
+# Usage:
+#   update-bundler <BUNDLER_VERSION>
+# Example:
+#   update-bundler 2.5.4
+#
+# This script will update Bundler to the specified version in all projects listed in ~/.config/bundler-projects.
+#
+# You will require Docker and GitHub CLI to run this script.
+
+set -e
+
+BUNDLER_VERSION="$1"
+if [ -z "$BUNDLER_VERSION" ]; then
+    echo "Usage: Please provide a Bundler version. For example:"
+    echo "update-bundler 2.5.4"
+    exit 1
+fi
+
+BUNDLER_RELEASE="https://github.com/rubygems/rubygems/releases/tag/bundler-v$BUNDLER_VERSION"
+
+run_in_container() {
+	container_id=$1
+	shift
+	cmd=$*
+
+	docker exec --workdir /app "$container_id" bash -c "$cmd"
+}
+
+while IFS= read -r folder
+do
+	cd "$(eval echo $folder)"
+
+	echo
+	echo "================================================="
+	echo "Processing: $folder"
+	echo "BUNDLER_VERSION: $BUNDLER_VERSION"
+	echo "BUNDLER_RELEASE: $BUNDLER_RELEASE"
+
+	ruby_version=$(ruby -e 'puts RUBY_VERSION')
+
+	docker pull ruby:${ruby_version}
+
+	git stash -u
+	git checkout main
+	git pull
+
+	container_id=$(docker run -d -v .:/app ruby:${ruby_version} sleep infinity)
+
+	echo "Bundling..."
+	run_in_container "$container_id" "bundle install"
+	echo "Installing new version of bundler..."
+	run_in_container "$container_id" "gem install bundler --version ${BUNDLER_VERSION}"
+	echo "Updating bundler..."
+	run_in_container "$container_id" "bundle _${BUNDLER_VERSION}_ update --bundler"
+
+	echo "Tidying up container..."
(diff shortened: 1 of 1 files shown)


smaboshe:
I am open for a discussion about whether this is the right place for a script like this to live 🙂.

nickcharlton:
We've been using this now in lots of projects (client and internal ones — the original motivation was on a project where we had four or five Rails apps), which I feel is quite good going for something I threw together in a few hours and then @smaboshe adjusted to make it work across several!

But, I've never been sure where it should or could live. I always felt like it'd be a cool little automated service, but it needed to start somewhere.

bin/update-bundler
```diff
@@ -0,0 +1,78 @@
...
+#
+# This script will update Bundler to the specified version in all projects listed in ~/.config/bundler-projects.
+#
+# You will require Docker and GitHub CLI to run this script.
+
+set -e
+
+BUNDLER_VERSION="$1"
+if [ -z "$BUNDLER_VERSION" ]; then
+    echo "Usage: Please provide a Bundler version. For example:"
```
mike-burns, line 18: Normally the "Usage:" line shows the usage.

```suggestion
    echo "Usage: update-bundler BUNDLER_VERSION"
```

bin/update-bundler
```diff
@@ -0,0 +1,78 @@
...
+
+run_in_container() {
+	container_id=$1
+	shift
+	cmd=$*
+
+	docker exec --workdir /app "$container_id" bash -c "$cmd"
+}
+
+while IFS= read -r folder
```
mike-burns, line 33: FWIW, "folder" is what the FAT filesystems call a directory. I like to use `directory` or `dir` for consistency with the rest of unix.

bin/update-bundler
```diff
@@ -0,0 +1,78 @@
...
+do
+	cd "$(eval echo $folder)"
```
mike-burns, line 35: What's `$folder`? I'm curious about the `eval echo`.

mike-burns (commented):
Seems useful.

This script does two things, and I wonder if it's worth separating them:

- run Docker commands over each directory, and
- upgrade Bundler

bin/update-bundler
smaboshe: `$folder` is the directory that is currently being worked on from the list defined in `~/.config/bundler-projects`.

I use symbolic links for some of the paths the scripts needs to navigate to. Without `eval`, `echo $folder` would not recognise these symbolic links as directory paths.

Would renaming `$folder` to `$working_directory` (or something like this) make this clearer?

smaboshe:
> This script does two things, and I wonder if it's worth separating them:

@mike-burns, I am open to this and am not sure where the boundaries would be. Happy to schedule some pairing time to explore this.

bin/update-bundler
mike-burns: Maybe readlink(1) is a more clear way to solve this (without introducing arbitrary evals).

```suggestion
	cd "$(readlink -f $folder)"
```

mike-burns:
Hm good point, line 51 is the first bottleneck. The two concepts are rather interwoven.

bin/update-bundler
smaboshe: Sounds good. I'll try this locally with `readlink -f` and see how it goes.

TIL!

nickcharlton:
I find it interesting how we perceive these different concepts, @mike-burns!

> run Docker commands over each directory, and
> upgrade Bundler

My sticking point (which is the version in my dotfiles repo) was that I thought the separation of concerns was rather:

1. Run the Bundle update script in a trusted clean environment (hence running in a container…),
2. Run this command over multiple projects

I didn't do 2, because I wasn't fully convinced it was in scope for 1 and then kind of got stuck.

bin/update-bundler
smaboshe: I am having challenges getting this to work locally using `readlink`. I think I may have my quotes or `$` setup incorrectly.

Would we want to replace `eval` at the end of the script with `readlink` as well?

purinkle (commented):
I lack the scripting knowledge to provide a code review, but I still have an opinion. 🤷‍♂️

This script adds a Docker dependency to our dotfiles. 🐳

We would need to add an entry into [`laptop`][] for compatibility. 🛠️

Do we now expect thoughtbot developers to install Docker as a default? 🤔

[`laptop`]: https://github.com/thoughtbot/laptop

bin/update-bundler
mike-burns: If `readlink` doesn't work then I guess `eval` is the right answer. That's too bad, but shell programming is like that.

mike-burns:
I see this as a tool for those who have to suffer under Docker, and not an endorsement for that annoying piece of software. I hope we don't add it to the laptop script!

That said, maybe we should rename this script to something that mentions Docker?

smaboshe:
> That said, maybe we should rename this script to something that mentions Docker?

I'm happy to!

bin/update-bundler
smaboshe: After learning more about `eval`, I hope I can get to a version of the script that works without it.

smaboshe:
Thank you, Everyone for the input on this. On reflection, I don't think that this iteration of the script is ready to be merged into `dotfiles`. I'll close this for now and hope to have a better version to work with before re-opening.

Sounds like there might be a separate discussion to be had about Docker as well?

Thanks again!

Issue #742, opened 2024-04-24
zsh function .zwc parse errors 


bengolder:
After installing, I get errors every time I open a new iTerm window:
```zsh
/Users/bengolder/.zsh/functions/_git_delete_branch.zwc:1: parse error near `)'
/Users/bengolder/.zsh/functions/change-extension.zwc:1: no such file or directory: ^G^F^E^D^A^H^A
/Users/bengolder/.zsh/functions/envup.zwc:1: no such file or directory: ^G^F^E^D^Ax^A
[3] 75334
/Users/bengolder/.zsh/functions/g.zwc:1: no such file or directory: ^G^F^E^D^A
[4] 75335
/Users/bengolder/.zsh/functions/g.zwc:1: no such file or directory:
[4]  + exit 127    ��
[3]  + exit 127    �A
/Users/bengolder/.zsh/functions/g.zwc:1: command not found: A
/Users/bengolder/.zsh/functions/mcd.zwc:1: no such file or directory: ^G^F^E^D^A\M-T
Wed Apr 24 10:42 AM ~ %
```

It's clear that these are occurring when the `zshrc` file runs this loop to source all the files in `.zsh/functions`:
```zsh
# load custom executable functions
for function in ~/.zsh/functions/*; do
  source $function
done
```

If I delete the `~/.zsh` directory and rerun `rcup`, the errors are absent until new `.zwc` files are autogenerated for each file in `~/.zsh/functions`:
```zsh
Wed Apr 24 10:49 AM functions % pwd
/Users/bengolder/.zsh/functions
Wed Apr 24 10:49 AM functions % ls -la
total 40
drwxr-xr-x  12 bengolder  staff  384 Apr 24 10:47 .
drwxr-xr-x   5 bengolder  staff  160 Apr 24 10:46 ..
lrwxr-xr-x   1 bengolder  staff   58 Apr 24 10:46 _git_delete_branch -> /Users/bengolder/dotfiles/zsh/functions/_git_delete_branch
-rw-r-----   1 bengolder  staff  480 Apr 24 10:47 _git_delete_branch.zwc
lrwxr-xr-x   1 bengolder  staff   56 Apr 24 10:46 change-extension -> /Users/bengolder/dotfiles/zsh/functions/change-extension
-rw-r-----   1 bengolder  staff  528 Apr 24 10:47 change-extension.zwc
[...]

PR #743, opened 2024-04-25
Skip zwc zsh function files

bengolder:
- Uses the `~` glob operator to be consistent with `_load_settings()``
- moved loading after `_load_settings``, because `setopt extendedglob` is needed for the `~` glob operator and is run in `_load_settings()`

### DIFF
--- a/zshrc
+++ b/zshrc
@@ -1,8 +1,3 @@
-# load custom executable functions
-for function in ~/.zsh/functions/*; do
-  source $function
-done
-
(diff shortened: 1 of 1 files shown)


bengolder:
Closes https://github.com/thoughtbot/dotfiles/issues/742

bengolder:
It seemed harmless and potentially only beneficial to move the loading of zsh functions to after `_load_settings`. I figured if anything, it allows the zsh functions to take advantage of any zsh configs, including the `setopt extendedglob`.

PR #756, opened 2024-11-20
Exclude CODEOWNERS file from symlinking

Hatlen:
I think this file wasn't meant to be symlinked (?). Nice that it was easy to figure out how to not link it.

Thanks for this great repo, learnt a ton from it :)


purinkle:
Great spot! Thank you for your contribution.

PR #758, opened 2025-02-06
ensure path.zsh works when asdf.sh file is missing in asdf v0.16.1

johnreitano:


purinkle (approved):
:woman_student: Looks good to me!

Issue #778, opened 2026-01-16
Newest version throws error in WSL ubuntu


maciejkowalski:
After update, when I start shell I have:

```
/home/hitsu/.zsh/configs/fzf.zsh:3: no such file or directory: /opt/homebrew/bin/fzf
/home/hitsu/.zshrc:37: no such file or directory: /opt/homebrew/bin/brew
~ %
```

I ran git blame and it points to this commit:

https://github.com/thoughtbot/dotfiles/commit/fd28b68b31a9af951dab6582058c9f38543f9a76

Not sure how to fix this.

Are `dotfiles` only supported by macs (OS X)? Previously I was using it on ubuntu (old xubuntu OR windows WSL + Ubuntu) without any errors...

PR #782, opened 2026-04-01
Detect Homebrew location for Intel support

### LINKED ISSUE #781: zshrc expecting wrong homebrew location
Since #767 there is a new line `eval "$(/opt/homebrew/bin/brew shellenv)"` which now on my x86_64 Mac will always error `no such file or directory: /opt/homebrew/bin/brew` because homebrew sits in `/usr/local`. There should be some check before running this command.

purinkle:
## Summary

- Fixes the hardcoded `/opt/homebrew/bin/brew` path that only works on Apple Silicon Macs
- Checks for Homebrew at `/opt/homebrew/bin/brew` (Apple Silicon) and `/usr/local/bin/brew` (Intel) before eval-ing `shellenv`

Closes #781

### DIFF
--- a/zshrc
+++ b/zshrc
@@ -34,7 +34,11 @@ _load_settings() {
 }
 _load_settings "$HOME/.zsh/configs"
 
-eval "$(/opt/homebrew/bin/brew shellenv)"
+if [ -x /opt/homebrew/bin/brew ]; then
+  eval "$(/opt/homebrew/bin/brew shellenv)"
+elif [ -x /usr/local/bin/brew ]; then
+  eval "$(/usr/local/bin/brew shellenv)"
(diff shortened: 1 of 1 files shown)


MatheusRich (approved):
LGTM. To confirm, are our dotfiles MacOs-only?

purinkle:
@MatheusRich, not macOS-only, but macOS is the only supported platform. `dotfiles` is a companion to `laptop`, which targets macOS on Apple Silicon and Intel. Someone raised the same question recently in #778. Linux users do use it, but we don't support that.

The `if` guard here means the line is a no-op on systems without homebrew rather than an error. That is a small improvement for those users, too.

